PR #112, opened 2017-06-12
Enable variant search in main page search box

zhx828:


zhx828:
@JiaoJiao123 no need to review variant.js/variant.scss/variant.html

app/index.html
```diff
@@ -52,19 +52,32 @@
...
+                    <span ng-repeat="alias in match.model.gene.geneAliases"><span
+                        ng-bind-html="alias | uibTypeaheadHighlight:query"></span><span ng-if="!$last">,</span>
               </span>
+                </div>
+            </div>
+            <div ng-switch-when="variant">
+                <div><span>{{match.model.gene.hugoSymbol}}</span> /
+                    <span
+                        ng-bind-html="(match.model.variant.name ? match.model.variant.name : match.model.variant.alteration) | uibTypeaheadHighlight:query"></span>
+                    <span style="float: right" ng-if="!match.model.variantExist">Not annotated by OncoKB</span></div>
```
zhx828: yes

JiaoJiao123: Then there is a problem. Since currently nothing will be displayed if the input has no matching.

zhx828:
I misunderstood your meaning. I thought you meant no match from API. No, it
won't be display if no match found. And this is for items which are not
annotated by OncoKB. You can only see it when the variant you queried is
not existed in OncoKB. For example BRAF K601M

app/styles/main.scss
```diff
@@ -197,13 +197,311 @@ a:hover, a:focus {
...
+            .no-annotation {
+                float: right
+            }
+        }
+
+        .oncogenic-icon-image.level1.oncogenic {
+            background-position: -13px -14px;
+        }
+
+        .oncogenic-icon-image.level2A.oncogenic {
```
JiaoJiao123: Could we assign the background-position systematically instead of manually like this? like using a for loop? 
It is hard to manage this way.

zhx828: We probably could. I borrowed the code from portal. I will try to do some refactoring in oncokb-client.js in the following weeks. But not in this pr.

app/scripts/controllers/gene.js
```diff
@@ -389,15 +389,17 @@ angular.module('oncokbStaticApp')
         $scope.meta = {};
         $scope.meta.clinicalTable = {};
         $scope.meta.biologicalTable = {};
+        $scope.meta.showGeneAddition = true;
+        $scope.meta.altFreFlag = true;
+        $scope.meta.highestLevels = [];
```
zhx828: Updated

PR #133, opened 2017-08-03
various fixes about variant page

JiaoJiao123:


app/scripts/services/utils.js
```diff
@@ -52,12 +52,13 @@ angular.module('oncokbStaticApp')
                     'NA') : 'NA';
             },
             getVariantCellContent: function(geneName, variantName) {
-                if (geneName && variantName && variantName.indexOf('/') === -1) {
-                    geneName = geneName.trim();
-                    variantName = variantName.trim();
-                    return '<a href="#/gene/' + geneName + '/variant/' + variantName + '">' + variantName + '</a>';
+                geneName = geneName.trim();
```
zhx828: Keep `if(geneName & variantName)`

app/scripts/services/utils.js
```diff
@@ -52,12 +52,13 @@ angular.module('oncokbStaticApp')
...
+                variantName = variantName.trim();
+                var formattedVariantName = variantName;
+                if (/\//.test(variantName)) {
+                    formattedVariantName = encodeURIComponent(encodeURIComponent(formattedVariantName));
```
zhx828: Why encodeURIComponent twice?

app/scripts/services/utils.js
```diff
@@ -52,12 +52,13 @@ angular.module('oncokbStaticApp')
...
                 }
-                return '<span>' + variantName + '</span>';
+                return '<a href="#/gene/' + geneName + '/variant/' + formattedVariantName + '">' + variantName + '</a>';
```
zhx828: If the geneName and variantName are invalid, return empty string

app/scripts/services/utils.js
JiaoJiao123: If you encode it only once, the server decodes it, then it's the same as not encoding it at all.

PR #169, opened 2018-01-11
Add search box on the menu

zhx828:
jan-11-2018 17-41-53


JiaoJiao123:
screen shot 2018-01-12 at 2 47 24 pm

JiaoJiao123:
@zhx828 After clicking one option in the returned list, the search box won't return to the default search icon until mouse over it again. It will remain the format as the image I posted. Could you make it go back to the original icon look after choosing a page?

Issue #202, opened 2018-07-18
Replace all wordings from variant to alteration


zhx828:
- Change the Actionable Variants in the Data Access page to Actionable Alterations
- Change All Variants to All Alterations

PR #204, opened 2018-07-31
Rearrange Actionable Genes

zhx828:
This solves #201
This relates to #202


app/scripts/app.js
```diff
@@ -57,11 +57,14 @@ angular
                 controller: 'GenesCtrl',
                 controllerAs: 'genes'
             })
-            .when('/gene/:geneName/variant/:variantName', {
+            .when('/gene/:geneName/alteration/:alterationName', {
                 templateUrl: 'views/gene.html',
                 controller: 'GeneCtrl',
                 controllerAs: 'gene'
             })
+            .when('/gene/:geneName/variant/:alterationName', {
```
victoria34: Why do you still keep variant in path?

zhx828: This is the old URL format. In portal, we are still referring to the /gene/:geneName/variant/:alterationName. I need to redirect that to the updated URL.

zhx828:
This pull request needs to go with release-0.3.1 in oncokb repo.

PR #286, opened 2019-07-02
Support annotate selected tumor type

zhx828:
Replace tumor type filter dropdown to a typeahead component
When user pastes content, automatically trigger the filtering

### DIFF
--- a/app/scripts/controllers/actionalGenes.js
+++ b/app/scripts/controllers/actionalGenes.js
@@ -3,6 +3,8 @@
 angular.module('oncokbStaticApp')
     .controller('actionableGenesCtrl', function($rootScope, $scope, $location,
                                                 $sce,
+                                                $q,
+                                                $timeout,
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/actionalGenes.js
```diff
@@ -121,9 +181,30 @@ angular.module('oncokbStaticApp')
...
+                        return !record.trim().endsWith("NOS");
+                    }).map(function(mainType) {
+                        return {
+                            type: 'Main Type',
+                            name: mainType
+                        };
+                    }));
+                    deferred.resolve();
+                }, function() {
+                    deferred.resolve();
```
victoria34: Did you make that on purpose? Should it be deferred.reject()?

zhx828: Yeah, I don't want to reject the call if the main type endpoint gives error. Just don't need to include the main types in the list.

PR #315, opened 2019-11-22
Fix collapse issue

victoria34:
Fix https://github.com/oncokb/oncokb/issues/1824


victoria34:
> I would not put into windowStore. That uses to manage all objects related to window. You can use AppStore. But I really think that logic should reside in header

I don't think we can solve it by putting the logic in header because React doesn't allow side affects. I think AppStore is used for global config. I'll create another store called `HeaderStore`.

PR #316, opened 2019-11-24
Use master/slave server redis as caching 

zhx828:
This solves https://github.com/oncokb/oncokb/issues/1729

### DIFF
--- a/src/main/resources/config/application-prod.yml
+++ b/src/main/resources/config/application-prod.yml
@@ -147,4 +147,7 @@ application:
   api-proxy-url: 'http://oncokb-core:8888'
   sitemap-enabled: true
   user-registration-webhook:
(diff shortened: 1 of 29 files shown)


src/main/resources/config/application-prod.yml
```diff
@@ -146,3 +146,8 @@ application:
   profile: prod
   api-proxy-url: 'http://oncokb-core:8888'
   sitemap-enabled: true
+  user-registration-webhook:
+  master-slave-redis-cache:
+    master-address: 'redis://oncokb-public-redis-master:6379'
+    slave-address: 'redis://oncokb-public-redis-slave:6379'
```
victoria34: Does it work that master and slave share the same port?

PR #324, opened 2019-12-02
Redirect to Home Page for 404 error

victoria34:
Fix https://github.com/oncokb/oncokb/issues/1831


victoria34:
@zhx828 Can we add `exact` for all routes before PageNotFound? Otherwise, `/login/abc` will direct to the login page.

zhx828:
@victoria34 for most of the pages, but not all. I will update and let you take a look.

victoria34:
> @victoria34 for most of the pages, but not all. I will update and let you take a look.

Thanks

PR #327, opened 2019-12-03
Create action to automatically build docker image

zhx828:


zhx828:
We should follow the changes in https://github.com/oncokb/oncokb/issues/1990

PR #340, opened 2019-12-12
Support PMIDLink

zhx828:


src/main/webapp/app/shared/utils/UrlUtils.tsx
```diff
@@ -68,31 +70,35 @@ export const MSILink: React.FunctionComponent<{}> = () => {
...
-        _.forEach(results, (matchedItem) => {
-          const datum = matchedItem;
-          switch ( regexp ) {
-            case REGEXP.PMID:
-              contentWithLink = contentWithLink.replace( new RegExp( `${datum}(?!s*,)`, 'g' ),
-                `<a href="${REGEXP_LINK[regexp]}${datum.split( ':' )[ 1 ].trim()}" target="_blank">${datum}</a>` );
-              break;
-            case REGEXP.NCTID:
-              contentWithLink = contentWithLink.replace( datum,
-                `<a href="${REGEXP_LINK[regexp]}${datum}" target="_blank">${datum}</a>` );
```
zhx828: @victoria34 we should avoid wrap element in string. It will prevent the typescript checks.

PR #361, opened 2020-01-03
Display all cbioportal mutations on mutation mapper

victoria34:
FIx https://github.com/oncokb/oncokb/issues/1881

### DIFF
--- a/src/main/webapp/app/store/AnnotationStore.ts
+++ b/src/main/webapp/app/store/AnnotationStore.ts
@@ -427,21 +427,36 @@ export class AnnotationStore {
     invoke: () => {
       // simply mapping by protein change, assuming that protein change is unique for alterations
       const indexedByProteinChange = _.keyBy(
-        this.mutationMapperDataPortal.result,
+        this.mutationMapperDataExternal.result,
         mutation => mutation.proteinChange
       );
 
-      const data = this.mutationMapperDataExternal.result
-        ? this.mutationMapperDataExternal.result.map(mutation => {
-            const portalMutation =
+      const data = this.mutationMapperDataPortal.result
+        ? this.mutationMapperDataPortal.result.map(mutation => {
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/store/AnnotationStore.ts
```diff
@@ -427,21 +427,36 @@ export class AnnotationStore {
...
+              : undefined;
+            const variantAllele = externalMutation
+              ? externalMutation.variantAllele
               : undefined;
 
             return {
-              ...mutation,
-              cancerType
+              gene: {
+                hugoGeneSymbol: this.gene.result.hugoSymbol
```
zhx828: Since you are using this.gene, so it needs to be waited in the await method

src/main/webapp/app/store/AnnotationStore.ts
```diff
@@ -427,21 +427,36 @@ export class AnnotationStore {
...
       const indexedByProteinChange = _.keyBy(
-        this.mutationMapperDataPortal.result,
+        this.mutationMapperDataExternal.result,
         mutation => mutation.proteinChange
       );
 
-      const data = this.mutationMapperDataExternal.result
-        ? this.mutationMapperDataExternal.result.map(mutation => {
-            const portalMutation =
+      const data = this.mutationMapperDataPortal.result
```
zhx828: No mutationMapperDataPortal.result check. It has default empty array anyway

src/main/webapp/app/store/AnnotationStore.ts
```diff
@@ -427,21 +427,36 @@ export class AnnotationStore {
...
+        ? this.mutationMapperDataPortal.result.map(mutation => {
+            const externalMutation =
```
zhx828: we don't need to initialize a variable since it's not used anywhere else

PR #372, opened 2020-01-31
Fixed OncoKB API issues

victoria34:
The previous version is wrong. The current version has been built on my local docker successfully.


victoria34:
The current yarn file is identical with the version of `b32ff06`(Update the version to 2.0.0)

PR #373, opened 2020-02-03
Show commercial users by default

victoria34:


src/main/webapp/app/pages/userManagement/UserDetailsPage.tsx
```diff
@@ -46,15 +56,46 @@ export default class UserDetailsPage extends React.Component<{
...
       notifyError(e, 'Error fetching users');
     }
   }
 
   @action
-  toggleFilter(query: any) {
-    this.currentSelectedFilter = query;
+  toggleFilter(button: string) {
+    this.currentSelectedButton = button;
+    switch (this.currentSelectedButton) {
```
zhx828: Do we need the switch? Seems like a if-else condition

victoria34: I think it's clear and easy for extending filters in the future and switch is also good for if-else conditions more than 3.

src/main/webapp/app/pages/userManagement/UserDetailsPage.tsx
```diff
@@ -85,7 +123,7 @@ export default class UserDetailsPage extends React.Component<{
   }
 
   private getStatus(activated: boolean) {
```
zhx828: This status func is dup from user management. I didn't compare these two files, there might be more. Let's don't worry about that yet until we figure out a plan for these two pages

src/main/webapp/app/pages/userManagement/UserDetailsPage.tsx
zhx828: I would not worry about the extension when the requirement is not clear yet. At this point they are dup code.

PR #385, opened 2020-02-27
Add email template: Send Intake Form

zhx828:
This solves https://github.com/oncokb/oncokb/issues/1911


victoria34:
Btw, JJ thinks we don't need the cc input box. I want to check with you first to ensure if I can remove it. If you also think we don't need that, I'll create a PR for front-end later. @zhx828

zhx828:
> Btw, JJ thinks we don't need the cc input box. I want to check with you first to ensure if I can remove it. If you also think we don't need that, I'll create a PR to remove that in front-end later. @zhx828

We use that to cc our license group when sending out the reply. But since we now sending out the intake form directly, I do not think the cc box is needed too. I will remove together in my pr.

PR #397, opened 2020-03-30
Fix bug when adamenic account is not in the clarification domains

zhx828:


src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -161,8 +164,13 @@ private TextObject getTextObject(String title, String content) {
...
         blocks.add(buildUserInfoBlock(user));
 
-        // Add Approve button
-        blocks.add(buildPlainTextBlock(ACADEMIC_CLARIFICATION_NOTE));
+        if(withClarificationNote) {
+            // Add clarification note
+            blocks.add(buildPlainTextBlock(ACADEMIC_CLARIFICATION_NOTE));
+        } else {
+            // Add Approve button
+            blocks.add(buildPlainTextBlock(COMMERCIAL_APPROVE_NOTE));
```
victoria34: Why we add commercial approve note in academic blocks

zhx828: @victoria34 updated, was a typo above causing the mistake.

PR #455, opened 2020-07-02
Update token usage by cronjob instead of basing on actually API calls

zhx828:
We should not update the token based on the api call, that will mess up with the cache.
This fixes https://github.com/oncokb/oncokb/issues/2095

This is related to https://github.com/oncokb/oncokb/issues/2110

### DIFF
--- a/.jhipster/TokenStats.json
+++ b/.jhipster/TokenStats.json
@@ -25,6 +25,13 @@
             "fieldValidateRules": [
                 "required"
             ]
+        },
+        {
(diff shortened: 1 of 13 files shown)


zhx828:
The pr needs to do more performance test. The current result is not idea. Running select statement with aggression takes about 10mins

zhx828:
There is no much of the improvement we can do for a select aggression statement. For 70 million records, the best performance we can do is 7mins after indexing. @jjgao maybe you have more insights on this?

PR #479, opened 2020-07-14
Using localStorage to get server api

Leowisd:
This implements oncokb/oncokb/#2134

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,37 +25,11 @@ Make sure your running environment is the following:
 
 ### Modify config
 
-`/pom.xml`
-
-You should confirm the database name, username and password under `liquibase-maven-plugin`
-
 `/src/main/resources/config/application-dev.yml`
 
 1. Confirm database name, username, password under `datasource` config.
 2. Change `api-proxy-url` to the URL where oncokb running. For example, `http://localhost:8888/oncokb`
 3. Make sure the password for your Redis as same as the password for Redis defined in this file
-
-**Modify the three client instances**
-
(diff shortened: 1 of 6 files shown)


src/main/webapp/app/shared/utils/DevUtils.ts
```diff
@@ -0,0 +1,9 @@
+import { LOCAL_DEV_OPT } from '../../config/constants';
+
+export function getClientInstanceURL(url: string) {
+  let devURL = localStorage.getItem(LOCAL_DEV_OPT) || '';
+
+  if (devURL.length > 0 && url.length > 0) devURL += '/';
+
+  return devURL + url;
```
zhx828, line 8:
```suggestion
if(devURL) {
 return url ? `${devURL}/${url}` : devURL;
} else {
  return url;
}
```

README.md
```diff
@@ -83,6 +57,22 @@ Add the `help` flag on any command to see how you can use it. For example, `yarn
 
 The `yarn run` command will list all of the scripts available to run for this project.
 
+### Connect with server side locally
+
+After starting project up locally, you should type the following command in your browser console
+
+```
+localStorage.setItem("localdev","PUBLIC_SERVER_API")
```
zhx828, line 65: I would not bother to allow user to specify the `PUBLIC_SERVER_API`. In practice, I assume most of us will use the 9095 as the public port. Just make localdev as a boolean value.
If there is a requirement, then we can add a new property to specify that.

Also in this pr, I don't see the default for PUBLIC_SERVER_API

Leowisd:
Close it. I'll submit a new pull request to fix.

PR #480, opened 2020-07-15
Using localStorage to set client api

Leowisd:
This implements https://github.com/oncokb/oncokb/issues/2134


zhx828:
@Leowisd nice work!

PR #526, opened 2020-09-30
Allow admin to change token expiration date

Leowisd:
This fix https://github.com/oncokb/oncokb/issues/2238


src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,178 @@
...
+} from 'app/config/constants';
+import MomentLocaleUtils from 'react-day-picker/moment';
+import { DefaultTooltip } from 'cbioportal-frontend-commons';
+
+type TokenExtendButtonProps = {
+  token: Token;
+};
+
+@observer
+export default class TokenExtendButton extends React.Component<
```
zhx828, line 32: Can we make this a general calendar component that pass in the current date and add an afterChangeDate method as property?

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,178 @@
...
+> {
+  @observable selectedDay: Date;
+
+  constructor(props: TokenExtendButtonProps) {
+    super(props);
+  }
+
+  @autobind
+  @action
+  extendExpirationDate(token: Token, newDate: string) {
```
zhx828, line 43: So basically this will just be a simple view component without data logic ties to it.

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,178 @@
...
+            style={{ textAlign: 'center' }}
+            onClick={() => {
+              this.extendExpirationDate(
+                this.props.token,
+                getMomentInstance(this.props.token.expiration)
+                  .add(1, 'y')
+                  .format(APP_LOCAL_DATETIME_FORMAT_Z_FORCE)
+              );
+            }}
+          >
```
zhx828, line 144: Can we make these dropdown items into a for loop to get rid of the dup code?

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,178 @@
...
+              }
+            }}
+            style={{ margin: '0 auto' }}
+            formatDate={MomentLocaleUtils.formatDate}
+            parseDate={MomentLocaleUtils.parseDate}
+            placeholder={`Expires on: ${getMomentInstance(
+              this.props.token.expiration
+            ).format(APP_LOCAL_DATE_FORMAT)}`}
+          />
+          <Row style={{ marginTop: '2px' }}>
```
zhx828, line 164: We don't modify margin this fashion anymore. we use `mt-1` in className. Check out https://getbootstrap.com/docs/4.5/utilities/spacing/

src/main/webapp/app/components/tokenInputGroups/TokenInputGroups.tsx
```diff
@@ -9,8 +9,10 @@ import classnames from 'classnames';
 import { observable, action } from 'mobx';
 import { observer } from 'mobx-react';
 import autobind from 'autobind-decorator';
+import TokenExtendButton from './TokenExtendButton';
 
 type TokenInputGroupsProps = {
+  extend: boolean;
```
zhx828, line 15: This is definitely not a good name. The purpose of this calendar is to change token expiration date. so maybe `changeTokenExpirationDate`?

src/main/webapp/app/config/constants.tsx
```diff
@@ -34,6 +34,7 @@ export const APP_TIMESTAMP_FORMAT = 'MM/DD/YY HH:mm:ss';
 export const APP_LOCAL_DATE_FORMAT = 'MM/DD/YYYY';
 export const APP_LOCAL_DATETIME_FORMAT = 'YYYY-MM-DDTHH:mm:ss';
 export const APP_LOCAL_DATETIME_FORMAT_Z = 'YYYY-MM-DDTHH:mm:ssZ';
+export const APP_LOCAL_DATETIME_FORMAT_Z_FORCE = 'YYYY-MM-DDTHH:MM:SS[Z]';
```
zhx828, line 37: why don't we use `APP_LOCAL_DATETIME_FORMAT_Z`?

zhx828 (changes requested):

src/main/webapp/app/config/constants.tsx
Leowisd: If we used `moment.add()` then `format(APP_LOCAL_DATETIME_FORMAT_Z)` it will not get this format but `YYYY-MM-DDTHH:mm:ss-H:00`. Here need to `[Z] `to convert the time zone `-H:00` on the tail.

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,156 @@
...
+import MomentLocaleUtils from 'react-day-picker/moment';
+import { DefaultTooltip } from 'cbioportal-frontend-commons';
+
+type TokenExtendButtonProps = {
+  currentDate: string;
+  afterChangeDate: (newDate: string) => void;
+};
+
+@observer
+export default class TokenExtendButton extends React.Component<
```
zhx828, line 30: After looking into the logics, I think this component is no longer needed to be a class. It could be a simple Functional component. Also the selectedDay no longer needs to be observable.

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,156 @@
...
+    this.props.afterChangeDate(
+      moment(this.selectedDay).format(APP_LOCAL_DATETIME_FORMAT_Z_FORCE)
+    );
+  }
+
+  render() {
+    const extendSchedule = [
+      {
+        amount: 1,
+        idx: 1,
```
zhx828, line 57: can we have a more meaningful key?

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,156 @@
...
+      },
+      {
+        amount: 1,
+        idx: 4,
+        unit: 'year'
+      }
+    ];
+    return (
+      <DefaultTooltip
+        placement={'top'}
```
zhx828, line 78: This is the default, no need to specify.

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,156 @@
...
+        overlay={'Extend token expiration date'}
```
zhx828, line 79: Let's say `Change token expiration date`

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,156 @@
...
+              case 'months':
+                unit = 'M';
+                break;
+              case 'year':
+              case 'years':
+                unit = 'y';
+                break;
+              default:
+                unit = 'd';
+                break;
```
zhx828, line 107: Why don't we just add one more property in the `extendSchedule`?

src/main/webapp/app/pages/AccountPage.tsx
```diff
@@ -231,8 +251,10 @@ export class AccountPage extends React.Component<IRegisterProps> {
               }
             >
               <TokenInputGroups
+                changeTokenExpirationDate={false}
                 tokens={this.tokens}
                 onDeleteToken={this.deleteToken}
+                extendExpirationDate={this.extendExpirationDate}
```
zhx828, line 257: We don't need this if the `changeTokenExpirationDate ` is false

zhx828 (changes requested):

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
Leowisd: If don't specify placement, it will pop up on the left side. Is that ok?

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
Leowisd: Cause the unit needs to be declared as `moment.unitOfTime.DurationConstructor` or it cannot be added by `Moment.add`.

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
zhx828: Hm that's weird. Let's keep it at the top. I took a look at the original default of the library. It must has been changed somewhere.

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,138 @@
...
+  function handleSelectedDay() {
+    props.afterChangeDate(
+      moment(selectedDay).format(APP_LOCAL_DATETIME_FORMAT_Z_FORCE)
+    );
+  }
+
+  const extendSchedule = [
+    {
+      amount: 1,
+      scheduleIndex: 1,
```
zhx828, line 42: I don't see the use case of adding this property. Can we just use the array index at the line 94?

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,138 @@
...
+    },
+    {
+      amount: 1,
+      scheduleIndex: 2,
+      unitText: 'month'
+    },
+    {
+      amount: 6,
+      scheduleIndex: 3,
+      unitText: 'months'
```
zhx828, line 53: No need to have two different unitText  because of the singular or plural. Use library `pluralize` which has been included, to pluralize the text.

src/main/webapp/app/components/tokenInputGroups/TokenExtendButton.tsx
```diff
@@ -0,0 +1,138 @@
...
+              unit = 'M';
+              break;
+            case 'year':
+            case 'years':
+              unit = 'y';
+              break;
+            default:
+              unit = 'd';
+              break;
+          }
```
zhx828, line 91: Again, I think you just need to include unit in the extendSchedule definition. No need to add a switch here.

src/main/webapp/app/pages/AccountPage.tsx
```diff
@@ -78,6 +78,26 @@ export class AccountPage extends React.Component<IRegisterProps> {
       });
   }
 
+  @action
+  extendExpirationDate(token: Token, newDate: string) {
+    client
+      .updateTokenUsingPUT({
+        token: {
+          ...token,
+          renewable: true,
```
zhx828, line 87: We should not change the renewable setting here.

src/main/webapp/app/pages/userPage/UserPage.tsx
```diff
@@ -148,6 +148,26 @@ export default class UserPage extends React.Component<IUserPage> {
       });
   }
 
+  @action
+  extendExpirationDate(token: Token, newDate: string) {
+    client
+      .updateTokenUsingPUT({
+        token: {
+          ...token,
+          renewable: true,
```
zhx828, line 157: Not here as well.

zhx828 (changes requested):

zhx828 (approved):

PR #532, opened 2020-10-02
Update to jhipster 6.9.1

zhx828:


src/main/java/org/mskcc/cbio/oncokb/config/StaticResourcesWebConfiguration.java
```diff
@@ -0,0 +1,49 @@
+package org.mskcc.cbio.oncokb.config;
```
zhx828, line 1: Need to revisit this file.

src/main/java/org/mskcc/cbio/oncokb/config/WebConfigurer.java
```diff
@@ -61,10 +61,7 @@ public void onStartup(ServletContext servletContext) throws ServletException {
         if (env.getActiveProfiles().length != 0) {
```
zhx828, line 61: Need to visit the file

PR #539, opened 2020-10-14
Fix redis cache & Update token extension calendar component 

zhx828:
Component update to:
Screen Shot 2020-10-14 at 5 32 13 PM

We also refetch the tokens after modifying so the new expiration date would be reflected.

This fixes https://github.com/oncokb/oncokb/issues/2246

### DIFF
--- a/src/main/webapp/app/components/calendarButton/CalendarButton.tsx
+++ b/src/main/webapp/app/components/calendarButton/CalendarButton.tsx
@@ -1,10 +1,13 @@
-import React from 'react';
+import React, { useState } from 'react';
 import {
   InputGroup,
   Button,
   Dropdown,
   DropdownButton,
   Row,
+  Form,
+  Col,
+  Container,
 } from 'react-bootstrap';
 import classnames from 'classnames';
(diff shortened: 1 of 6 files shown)


src/main/webapp/app/components/calendarButton/CalendarButton.tsx
```diff
@@ -24,7 +27,14 @@ type CalendarButtonProps = {
   afterChangeDate: (newDate: string) => void;
 };
 
+enum ExtensionType {
+  FROM_TODAY = 'Today',
+  FROM_EXPIRATION = 'Expiration Day',
+}
+
 export const CalendarButton: React.FunctionComponent<CalendarButtonProps> = props => {
+  const [extensionType, setExtensionType] = useState(ExtensionType.FROM_TODAY);
```
zhx828, line 36: In order to continue using FunctionalComponent, I use a react hook here.

zhx828:
@Leowisd remember hide the whitespaces when doing the review. There are some indentations added.

Leowisd (approved):

PR #550, opened 2020-11-17
Send email when token exposed

Leowisd:
Implements https://github.com/oncokb/oncokb/issues/2298

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
@@ -19,6 +19,9 @@ public enum MailType {
     , VERIFY_EMAIL_BEFORE_ACCOUNT_EXPIRES("verifyEmailBeforeAccountExpires", "Verify user still owns the email address", null)
     , APPROVAL_MSK_IN_COMMERCIAL("approvalMSKInCommercial", "Autocorrect MSK user license to academic", null)
     , TRIAL_ACCOUNT_IS_ABOUT_TO_EXPIRE("trialAccountIsAboutToExpire", "Trail account is about to expire", null)
+    , TOKEN_HAS_BEEN_EXPOSED("tokenHasBeenExposed", "Token has been exposed", null)
+    , TOKEN_HAS_BEEN_EXPOSED_USER("tokenHasBeenExposedToUser", "Token has been exposed", null)
+    , SEARCHING_RESPONSE_STRUCTURE_HAS_CHANGED("searchingResponseStructureHasChanged", "Searching Response Structure Has Changed", null)
     , TEST("testEmail", "Test", null)
     ;
 
--- a/src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
@@ -11,6 +11,7 @@ import org.mskcc.cbio.oncokb.domain.enumeration.LicenseType;
 import org.mskcc.cbio.oncokb.domain.enumeration.MailType;
 import org.mskcc.cbio.oncokb.service.dto.UserDTO;
 import org.mskcc.cbio.oncokb.service.dto.UserMailsDTO;
+import org.mskcc.cbio.oncokb.web.rest.vm.ExposedToken;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.MessageSource;
@@ -32,6 +33,9 @@ import java.util.stream.Collectors;
 
 import static org.mskcc.cbio.oncokb.config.Constants.MAIL_LICENSE;
 import static org.mskcc.cbio.oncokb.domain.enumeration.MailType.TRIAL_ACCOUNT_IS_ABOUT_TO_EXPIRE;
+import static org.mskcc.cbio.oncokb.domain.enumeration.MailType.TOKEN_HAS_BEEN_EXPOSED;
+import static org.mskcc.cbio.oncokb.domain.enumeration.MailType.TOKEN_HAS_BEEN_EXPOSED_USER;
+import static org.mskcc.cbio.oncokb.domain.enumeration.MailType.SEARCHING_RESPONSE_STRUCTURE_HAS_CHANGED;
 
 /**
  * Service for sending emails.
@@ -234,6 +238,45 @@ public class MailService {
         } catch (MailException | MessagingException e) {
             log.warn("Email could not be sent to user '{}'", applicationProperties.getEmailAddresses().getTechDevAddress(), e);
         }
+    }
+
+    @Async
+    public void sendExposedTokensInfoMail(List<ExposedToken> tokens){
+        Context context = new Context(Locale.US);
+        context.setVariable("tokens", tokens);
+        context.setVariable(BASE_URL, jHipsterProperties.getMail().getBaseUrl());
+        String content = templateEngine.process("mail/" + TOKEN_HAS_BEEN_EXPOSED.getTemplateName(), context);
+
+        try{
+            sendEmail(applicationProperties.getEmailAddresses().getTechDevAddress(), applicationProperties.getEmailAddresses().getTechDevAddress(), null, "Exposed tokens", content, null, false, true);
+            log.info("Sent email to User '{}'", applicationProperties.getEmailAddresses().getTechDevAddress());
+        }
+        catch (MailException | MessagingException e){
+            log.warn("Email could not be sent to user '{}'", applicationProperties.getEmailAddresses().getTechDevAddress(), e);
+        }
+    }
+
+    @Async
+    public void sendMailWhenSearchingStructrueChange(String source){
+        Context context = new Context(Locale.US);
+        context.setVariable("source", source);
+        context.setVariable(BASE_URL, jHipsterProperties.getMail().getBaseUrl());
+        String content = templateEngine.process("mail/" + SEARCHING_RESPONSE_STRUCTURE_HAS_CHANGED.getTemplateName(), context);
+        try{
+            sendEmail(applicationProperties.getEmailAddresses().getTechDevAddress(), applicationProperties.getEmailAddresses().getTechDevAddress(), null, "Searching Response Structure Has Changed", content, null, false, true);
+            log.info("Sent email to User '{}'", applicationProperties.getEmailAddresses().getTechDevAddress());
+        }
+        catch (MailException | MessagingException e){
+            log.warn("Email could not be sent to user '{}'", applicationProperties.getEmailAddresses().getTechDevAddress(), e);
+        }
+    }
+
+    @Async
+    public void sendMailToUserWhenTokenExposed(UserDTO user, ExposedToken token){
+        Context context = new Context(Locale.US);
+        context.setVariable("token", token);
+        sendEmailFromTemplate(user, MailType.TOKEN_HAS_BEEN_EXPOSED_USER, "OncoKB Token exposed", 
+        applicationProperties.getEmailAddresses().getRegistrationAddress(), applicationProperties.getEmailAddresses().getTechDevAddress(), null, context);
     }
 
     public MailType getIntakeFormMailType(LicenseType licenseType) {
--- a/src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
@@ -1,8 +1,8 @@
 package org.mskcc.cbio.oncokb.web.rest;
 
-
 import io.github.jhipster.security.RandomUtil;
 import org.apache.commons.lang3.StringUtils;
+import org.mskcc.cbio.oncokb.config.application.ApplicationProperties;
 import org.mskcc.cbio.oncokb.domain.Token;
 import org.mskcc.cbio.oncokb.domain.User;
 import org.mskcc.cbio.oncokb.domain.enumeration.MailType;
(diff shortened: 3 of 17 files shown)


src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -182,6 +205,97 @@ public void checkTrialAccounts() {
...
+            if (githubCount > 0 || googleCount > 0 || baiduCount > 0){
+                ExposedToken t = new ExposedToken();
+                UserDTO user = userMapper.userToUserDTO(token.getUser());
+                t.setToken(q);
+                t.setEmail(user.getEmail());
+                t.setFirstName(user.getFirstName());
+                t.setLastName(user.getLastName());
+                t.setLicenseType(user.getLicenseType() != null ? user.getLicenseType().name() : "");
+                String source = "";
+                if(githubCount > 0){
```
zhx828, line 265: This is optional, I would just do a List and do a collection joining later.

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -182,6 +205,97 @@ public void checkTrialAccounts() {
...
+            }
+
+            int googleCount = 0;
+            try{
+                HttpEntity entity = getResponse(q).getEntity();
+                Document googleDoc = Jsoup.parse(EntityUtils.toString(entity));
+                Elements resultEle = googleDoc.select("h3");
+                googleCount = resultEle.size();
+                if (googleCount > 0) {
+                    String description = googleDoc.select(".kCrYT > div > div > div > div > div").text();
```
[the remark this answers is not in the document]
Leowisd: I think it's okay for now. I found the structure from four years ago also looks like that.

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
Leowisd: Could you please show me how to do it here? I've tried but I cannot implement same result by using such process. Appreciate!

zhx828: @Leowisd do a ArrayList, then add element if applicable. Then do a .stream().collect(Collectors.joining(" "))

Leowisd: Oh, I got it. I misunderstood your first comment. Sorry about that.

Leowisd:
@zhx828 Have updated.

zhx828 (changes requested):
@Leowisd let's update this pr with the following

- [x] Disable the token right away when a token is identified exposed, generate a new one for the user with other fields copied(expiration/usage/limit etc.), and delete the previous token

- [ ] Send an email to user stating the reason why we update their token. And give them reminder that do not expose the token in public access repositories

- [ ] Double check the a email record is added after sending to the user. This is useful when we later to analysis how many times the user exposes the token

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -182,6 +205,105 @@ public void checkTrialAccounts() {
...
+            }
+            sleep(1000);
+        }
+        log.info("Searching exposed tokens pipeline finished!");
+        if (results.size() > 0){
+            mailService.sendExposedTokensInfoMail(results);
+        }
+    }
+
+    private HttpResponse getResponse(String query) {
```
zhx828, line 287: Let's rename to getGoogleResponse

zhx828 (changes requested):
Looking good! For these two emails, could you send an example emails to our dev.oncokb@gmail.com? I will ask the team to review.

src/main/resources/templates/mail/tokenHasBeenExposed.html
```diff
@@ -0,0 +1,45 @@
...
+              <td th:text="${token.email}"/>
+              <td th:text="${token.firstName}"/>
+              <td th:text="${token.lastName}"/>
+            </tr>
+          </tbody>
+        </table>
+        <p>
+          New tokens for those users have been generated.
+        </p>
+        <p>
```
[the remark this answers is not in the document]
Leowisd: I'm not really understand about this issue. Have we already added
```
Sincerely,
The OncoKB Team
```
as the signature?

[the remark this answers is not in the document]
zhx828: @Leowisd in the message.properties, you have the email.signature can be used. Here is an example https://github.com/oncokb/oncokb-public/blob/master/src/main/resources/templates/mail/activationEmail.html#L24. But I also like to consolidate the `#{email.activation.text2}`. In varies templates, we use different properties which should just be one.

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -182,6 +205,109 @@ public void checkTrialAccounts() {
...
+          e.printStackTrace();
+        }
+        return result;
+    }
+
+    private void updateExposedToken(Token token){
+        Token newToken = tokenProvider.createToken(token.getUser(), Optional.of(token.getExpiration()), Optional.of(token.isRenewable()));
+        token.setToken(newToken.getToken());
+        tokenService.save(token);
+        tokenService.delete(newToken.getId());
```
zhx828, line 308: I'm bit confused here. Save the original token and delete the new one?

Leowisd: Here I updated the UUID of the original token with the new one. I tried copy information from old one to new onw and delete the old one directly, but it failed because the original token ID is foreign key of other schemes. If we wanna delete it, we should change all related data. In this case, I only updated the original token UUID with the new token and deleted the new token.

zhx828: @Leowisd yeah, it's associated with the token stats. Let's create a new one and set the expiration of the old one to the time of running this script.

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
```diff
@@ -19,6 +19,8 @@
     , VERIFY_EMAIL_BEFORE_ACCOUNT_EXPIRES("verifyEmailBeforeAccountExpires", "Verify user still owns the email address", null)
     , APPROVAL_MSK_IN_COMMERCIAL("approvalMSKInCommercial", "Autocorrect MSK user license to academic", null)
     , TRIAL_ACCOUNT_IS_ABOUT_TO_EXPIRE("trialAccountIsAboutToExpire", "Trail account is about to expire", null)
+    , TOKEN_HAS_BEEN_EXPOSED("tokenHasBeenExposed", "Token has been exposed", null)
+    , TOKEN_HAS_BEEN_EXPOSED_USER("tokenHasBeenExposed2User", "Token has been exposed", null)
```
zhx828, line 23: Let's keep the consistency to use `To`
```suggestion
    , TOKEN_HAS_BEEN_EXPOSED_USER("tokenHasBeenExposedToUser", "Token has been exposed", null)
```

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
```diff
@@ -236,6 +239,30 @@ public void sendTrialAccountExpiresMail(int expiresInDays, List<UserDTO> users)
...
+            sendEmail(applicationProperties.getEmailAddresses().getTechDevAddress(), applicationProperties.getEmailAddresses().getTechDevAddress(), null, "Exposed tokens", content, null, false, true);
+            log.info("Sent email to User '{}'", applicationProperties.getEmailAddresses().getTechDevAddress());
+        }
+        catch (MailException | MessagingException e){
+            log.warn("Email could not be sent to user '{}'", applicationProperties.getEmailAddresses().getTechDevAddress(), e);
+        }
+    }
+
+    @Async
+    public void sendExposedTokensInfoMailToUser(UserDTO user, ExposedToken token){
```
zhx828, line 259: Let's make this name more clear since it's quite long
```suggestion
    public void sendMailToUserWhenTokenExposed(UserDTO user, ExposedToken token){
```

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -182,6 +205,112 @@ public void checkTrialAccounts() {
         mailService.sendTrialAccountExpiresMail(DAYS_TO_CHECK, userDTOS);
     }
 
+    @GetMapping(path = "/check-exposed-tokens")
+    public void exposedTokensInfo() throws IOException, InterruptedException {
```
zhx828, line 209: Let's keep the method name same as the api url.
```suggestion
    public void checkExposedTokens() throws IOException, InterruptedException {
```

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -182,6 +205,112 @@ public void checkTrialAccounts() {
...
+        List<ExposedToken> results = new ArrayList<>();
+        for (Token token : tokens){
+            String q = token.getToken().toString();
+
+            int githubCount = 0;
+            try {
+                GitHub github = new GitHubBuilder().withOAuthToken(applicationProperties.getGithubToken()).build();
+                List<GHContent> gitRes = github.searchContent().q(q).list().toList();
+                githubCount = gitRes.size();
+                if (githubCount == 1 && gitRes.get(0).getHtmlUrl().indexOf("https://github.com/knowledgesystems/oncokb-deployment/") == 0){
```
zhx828, line 225: What's this? Why you specify the oncokb-deployment?

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -182,6 +205,112 @@ public void checkTrialAccounts() {
...
+            }
+
+            int googleCount = 0;
+            try{
+                HttpEntity entity = getGoogleResponse(q).getEntity();
+                Document googleDoc = Jsoup.parse(EntityUtils.toString(entity));
+                Elements resultEle = googleDoc.select("h3");
+                googleCount = resultEle.size();
+                if (googleCount > 0) {
+                    String description = googleDoc.select(".kCrYT > div > div > div > div > div").text();
```
zhx828, line 240: I suppose this structure will change in the future. What's our plan if that happens?

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -182,6 +205,112 @@ public void checkTrialAccounts() {
...
+                }
+            }
+            catch (Exception e){
+                e.printStackTrace();
+            }
+            
+            int baiduCount = 0;
+            try{
+                Document document = Jsoup.connect(String.format("https://www.baidu.com/s?wd=\"%s\"", q)).get();
+                Elements elements = document.select("div.result.c-container");
```
zhx828, line 252: Here too. What's the plan?

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -182,6 +205,112 @@ public void checkTrialAccounts() {
...
+    }
+
+    private void updateExposedToken(Token token){
+        Token newToken = tokenProvider.createToken(token.getUser(), Optional.of(token.getExpiration()), Optional.of(token.isRenewable()));
+        newToken.setCreation(token.getCreation());
+        newToken.setCurrentUsage(token.getCurrentUsage());
+        newToken.setUsageLimit(token.getUsageLimit());
+        token.setExpiration(Instant.now());
+        tokenService.save(token);
+        tokenService.save(newToken);
```
zhx828, line 311: Let's create a new createToken  method in the tokenProvider to take a Token as a input to avoid the multiple save.

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
Leowisd: Because there are some tokens in `/knowledgesystems/oncokb-deployment/` configmaps. If I used my personal github token to check, it will find oncokb tokens in that repo exposed since my github account has the permission to visit this private repo. If our dev github account doesn't have the permission to visit such private repo, then we don't need this.

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
Leowisd: For google, I have tried its APIs console, but I didn't find a good way to get the same searching result like using google searching. Maybe we could try again in the future to see if I missed some things. But for Baidu, I didn't find it provides any APIs for us to searching. Maybe we can only keep eyes on it or do not index exposed tokens on Baidu. If those structures changed in the future, and there are not better solutions to solve, maybe we can only index tokens on Github.

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
zhx828: @Leowisd let's remove this clause. oncokbdev account only has the access to the oncokb organization which only contains public repo.

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
zhx828: @Leowisd in this case, do you think you can first identify this section exists? If it does not, send an email to the dev team so we can investigate?

Leowisd:
@zhx828 Updated.

zhx828 (approved):

zhx828:
@Leowisd could you do a squash?

PR #557, opened 2020-12-03
Update Level of Envidence Page to display TX/DX/PX seperately

Leowisd:
Implemented https://github.com/oncokb/oncokb/issues/2346

chrome-capture


src/main/webapp/app/pages/LevelOfEvidencePage.tsx
```diff
@@ -111,11 +137,12 @@ export default class LevelOfEvidencePage extends React.Component<
...
 
   render() {
-    return (
-      <DocumentTitle title={DOCUMENT_TITLES.LEVELS}>
-        <div className="levels-of-evidence">
-          <>
-            <Row>
+    const tabs = [];
+    for (const version in Version) {
+      if (version !== Version.V1 && version !== Version.AAC) {
```
zhx828, line 142: This is error prone. Let's only loop the versions that need to have a tab.

zhx828 (changes requested):
One minor comment, nice work

zhx828 (approved):

PR #564, opened 2020-12-09
update genes page and variant page to show Dx/Px

Leowisd:
This implemented https://github.com/oncokb/oncokb/issues/2352

Mockup:

- Gene Page

chrome-capture

- Variant Page

chrome-capture (1)

### DIFF
--- a/src/main/webapp/app/pages/annotationPage/AnnotationPage.tsx
+++ b/src/main/webapp/app/pages/annotationPage/AnnotationPage.tsx
@@ -4,14 +4,14 @@ import { GenePageLink } from 'app/shared/utils/UrlUtils';
 import {
   DEFAULT_MARGIN_BOTTOM_LG,
   EVIDENCE_TYPES,
+  LEVEL_TYPES,
+  LEVEL_TYPE_NAMES,
   TABLE_COLUMN_KEY,
   TREATMENT_EVIDENCE_TYPES,
 } from 'app/config/constants';
 import styles from 'app/pages/alterationPage/AlterationPage.module.scss';
 import InfoIcon from 'app/shared/icons/InfoIcon';
-import OncoKBTable, {
-  SearchColumn,
-} from 'app/components/oncokbTable/OncoKBTable';
+import { SearchColumn } from 'app/components/oncokbTable/OncoKBTable';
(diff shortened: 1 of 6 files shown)


src/main/webapp/app/pages/annotationPage/AnnotationPage.tsx
```diff
@@ -227,6 +336,82 @@ export default class AnnotationPage extends React.Component<IAnnotationPage> {
...
+    return <div>{this.getTable(key)}</div>;
+  }
+
+  @computed
+  get tabs() {
+    const tabs: { title: string; key: LEVEL_TYPES }[] = [];
+    if (this.therapeuticImplications.length > 0) {
+      tabs.push({
+        key: LEVEL_TYPES.TX,
+        title: `Annotations with ${LEVEL_TYPE_NAMES.Tx} Implications`,
```
zhx828, line 390: Let's just call it Therapeutic Implications. Could you also update below?

src/main/webapp/app/shared/utils/Utils.tsx
```diff
@@ -272,7 +272,7 @@ export function getDefaultColumnDefinition<T>(
         Header: <span>Tumor Type</span>,
         accessor: 'cancerType',
         style: { whiteSpace: 'normal' },
-        minWidth: 150,
+        minWidth: 250,
```
zhx828, line 275: Why are we changing the minWidth?

zhx828 (changes requested):
Few minor comments. Looks good.

src/main/webapp/app/shared/utils/Utils.tsx
Leowisd: Because some dx/px tumor types are too long to be displayed in the table column. For example, `B-Lymphoblastic Leukemia/Lymphoma with t(9;22)(q34.1;q11.2);BCR-ABL1` will be displayed as two lines. It looks not good. Changing minWidth here could make every row in table has the same format.

zhx828: Overwrite the minWidth at the place you use this column definition.

zhx828 (approved):

PR #565, opened 2020-12-11
update recaptcha - ignore home page

Leowisd:
This implements https://github.com/oncokb/oncokb/issues/2361

 Disable recaptcha on home page can reduce the possibility of showing challenge in incognito model.

### DIFF
--- a/src/main/webapp/app/shared/auth/RecaptchaBoundaryRoute.tsx
+++ b/src/main/webapp/app/shared/auth/RecaptchaBoundaryRoute.tsx
@@ -1,31 +1,51 @@
-import { RECAPTCHA_KEY } from 'app/indexUtils';
 import ErrorBoundaryRoute from 'app/shared/error/error-boundary-route';
+import WindowStore from 'app/store/WindowStore';
+import { computed } from 'mobx';
+import { observer } from 'mobx-react';
 import React from 'react';
 import { RouteProps } from 'react-router-dom';
 import { ContactLink } from '../links/ContactLink';
 
 export interface IRecaptchaBoundaryRoute extends RouteProps {
   isUserAuthenticated: boolean;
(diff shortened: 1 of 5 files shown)


src/main/webapp/app/shared/auth/RecaptchaBoundaryRoute.tsx
```diff
@@ -1,31 +1,51 @@
...
-          .
-        </div>
-      </>
-    );
+@observer
+export class RecaptchaBoundaryRoute extends React.Component<
+  IRecaptchaBoundaryRoute
+> {
+  @computed
+  get recaptchaVerifyed() {
```
zhx828, line 19: Why are we computing on a variable?

src/main/webapp/app/store/WindowStore.ts
```diff
@@ -11,6 +11,7 @@ export interface IWindowSize {
 class WindowStore {
   @observable size: IWindowSize;
   public recaptchaRef: any;
+  @observable recaptchaVerified: boolean;
```
zhx828, line 14: The goal is to verify one time and works for the rest of the session?

zhx828 (changes requested):

src/main/webapp/app/shared/auth/RecaptchaBoundaryRoute.tsx
Leowisd: This route component should be re-render to show real page component when the first time recaptcha got verified, or the page will only show warning message.

src/main/webapp/app/store/WindowStore.ts
Leowisd: Yes.

zhx828 (approved):

PR #566, opened 2020-12-11
Update response for Google/Baidu checking

Leowisd:
This implemented https://github.com/oncokb/oncokb/issues/2362

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
@@ -241,10 +241,10 @@ public class MailService {
     }
 
     @Async
-    public void sendExposedTokensInfoMail(List<ExposedToken> tokens, List<ExposedToken> toCheck){
+    public void sendExposedTokensInfoMail(List<ExposedToken> exposedTokens, List<ExposedToken> tokensToVerify){
         Context context = new Context(Locale.US);
-        context.setVariable("tokens", tokens);
-        context.setVariable("toCheck", toCheck);
+        context.setVariable("exposedTokens", exposedTokens);
+        context.setVariable("tokensToVerify", tokensToVerify);
         context.setVariable(BASE_URL, jHipsterProperties.getMail().getBaseUrl());
         String content = templateEngine.process("mail/" + TOKEN_HAS_BEEN_EXPOSED.getTemplateName(), context);
 
--- a/src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
@@ -241,9 +241,10 @@ public class MailService {
     }
 
     @Async
-    public void sendExposedTokensInfoMail(List<ExposedToken> tokens){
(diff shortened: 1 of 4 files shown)


src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -214,6 +214,7 @@ public void checkExposedTokens() throws IOException, InterruptedException {
                                 .collect(Collectors.toList());
         log.info("Searching exposed tokens pipeline begins...");
         List<ExposedToken> results = new ArrayList<>();
+        List<ExposedToken> checkResults = new ArrayList<>();
```
zhx828, line 217: Let's rename to `unverifiedResults`

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -261,7 +262,21 @@ public void checkExposedTokens() throws IOException, InterruptedException {
...
-            if (githubCount > 0 || googleCount > 0 || baiduCount > 0){   
+            if (githubCount > 0){   
+                ExposedToken t = new ExposedToken();
+                UserDTO user = userMapper.userToUserDTO(token.getUser());
+                t.setToken(q);
+                t.setEmail(user.getEmail());
+                t.setFirstName(user.getFirstName());
+                t.setLastName(user.getLastName());
+                t.setLicenseType(user.getLicenseType() != null ? user.getLicenseType().name() : "");
+                t.setSource("GitHub");
```
zhx828, line 273: I believe this part of the logics are duplication of the below. It would be better to unify.

src/main/resources/templates/mail/tokenHasBeenExposed.html
```diff
@@ -35,7 +35,36 @@
...
+          <tbody>
+            <tr th:each="token: ${toCheck}">
+              <td th:text="${token.token}"/>
+              <td th:text="${token.source}"/>
+              <td th:text="${token.licenseType}"/>
+              <td th:text="${token.email}"/>
+              <td th:text="${token.firstName}"/>
+              <td th:text="${token.lastName}"/>
+            </tr>
+          </tbody>
```
zhx828, line 64: Should you only include when `toCheck` is not empty?

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
```diff
@@ -241,9 +241,10 @@ public void sendTrialAccountExpiresMail(int expiresInDays, List<UserDTO> users)
     }
 
     @Async
-    public void sendExposedTokensInfoMail(List<ExposedToken> tokens){
+    public void sendExposedTokensInfoMail(List<ExposedToken> tokens, List<ExposedToken> toCheck){
```
zhx828, line 244: Let's clarify the names. Maybe `exposedTokens` and `tokensToVerify`

zhx828 (changes requested):
Let's make the solution more generic? Essentially, we want to differentiate the confirmed exposed vs exposure to be verified

zhx828 (approved):

PR #573, opened 2020-12-17
Fix URL paramater bug

Leowisd:
This fixed https://github.com/oncokb/oncokb/issues/2370

The reason of item one and item two in this issue is we set those pages' routes as private route in current instance.

### DIFF
--- a/src/main/webapp/app/pages/genePage/GenePage.tsx
+++ b/src/main/webapp/app/pages/genePage/GenePage.tsx
@@ -15,6 +15,7 @@ import { Gene } from 'app/shared/api/generated/OncoKbAPI';
 import styles from './GenePage.module.scss';
 import {
   filterByKeyword,
+  generateLoELink,
   getCancerTypeNameFromOncoTreeType,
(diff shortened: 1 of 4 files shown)


src/main/webapp/app/pages/genePage/GenePage.tsx
```diff
@@ -558,21 +561,39 @@ export default class GenePage extends React.Component<GenePageProps> {
...
     } else if (key === TAB_KEYS.DX) {
       return (
         <span>
           A list of diagnostic {this.store.hugoSymbol} alterations and the
-          corresponding OncoKB diagnostic level of evidence.
+          corresponding
+          <Link
+            to={`${PAGE_ROUTE.LEVELS}#${QueryString.stringify({
+              version: LEVEL_TYPES.DX.toUpperCase(),
+            })}`}
```
zhx828, line 575: Can we write an util to generate the level page link?

zhx828 (changes requested):

zhx828 (approved):

PR #583, opened 2021-01-06
Update to the latest api client

zhx828:
I believe these changes were caused by the recent usage analysis changes.


src/main/java/org/mskcc/cbio/oncokb/web/rest/vm/usageAnalysis/UsageSummary.java
```diff
@@ -1,20 +1,22 @@
...
+
 import java.util.Map;
 
 /**
  * Created by Yifu Yao on 2020-10-28
  */
 
 public class UsageSummary {
-    private Map<String, Map<String, Integer>> month;
+    private Map<String, JSONObject> month;
```
zhx828, line 12: Swagger generator cannot map nested map properly. It will generate a weird characters which causes typescript error.

Leowisd (approved):

PR #603, opened 2021-01-15
Implement user usage analysis pipeline

Leowisd:
Implemented https://github.com/oncokb/oncokb/issues/2349

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/querydomain/UserTokenUsageWithInfo.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/querydomain/UserTokenUsageWithInfo.java
@@ -1,0 +1,14 @@
+package org.mskcc.cbio.oncokb.querydomain;
+
+import org.mskcc.cbio.oncokb.domain.Token;
+
+/**
+ * Created by Yifu Yao on 2021-1-14
+ */
+
+public interface UserTokenUsageWithInfo {
+    Integer getCount();
+    Token getToken();
+    String getTime();
+    String getResource();
+}
--- a/src/main/java/org/mskcc/cbio/oncokb/repository/TokenStatsRepository.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/repository/TokenStatsRepository.java
@@ -3,6 +3,7 @@ package org.mskcc.cbio.oncokb.repository;
 import org.mskcc.cbio.oncokb.domain.TokenStats;
 
 import org.mskcc.cbio.oncokb.querydomain.UserTokenUsage;
+import org.mskcc.cbio.oncokb.querydomain.UserTokenUsageWithInfo;
 import org.springframework.data.jpa.repository.*;
 import org.springframework.stereotype.Repository;
 
@@ -19,4 +20,10 @@ public interface TokenStatsRepository extends JpaRepository<TokenStats, Long> {
 
     @Query("select sum(tokenStats.usageCount) as count, tokenStats.token as token from TokenStats tokenStats where tokenStats.accessTime < ?1 group by tokenStats.token")
     List<UserTokenUsage> countTokenUsageByToken(Instant before);
+
+    @Query("select sum(tokenStats.usageCount) as count, tokenStats.token as token, DATE_FORMAT(tokenStats.accessTime,'%Y-%m') as time, tokenStats.resource as resource " + 
+    " from TokenStats tokenStats " + 
+    " where tokenStats.accessTime > ?1 " + 
+    " group by tokenStats.token, DATE_FORMAT(tokenStats.accessTime,'%Y-%m'), tokenStats.resource")
+    List<UserTokenUsageWithInfo> countTokenUsageByTokenTimeResource(Instant after);
 }
--- a/src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
@@ -5,8 +5,11 @@ import org.apache.commons.lang3.StringUtils;
 import org.mskcc.cbio.oncokb.config.application.ApplicationProperties;
 import org.mskcc.cbio.oncokb.domain.Token;
 import org.mskcc.cbio.oncokb.domain.User;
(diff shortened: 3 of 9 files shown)


src/main/java/org/mskcc/cbio/oncokb/repository/TokenStatsRepository.java
```diff
@@ -19,4 +20,10 @@
 
     @Query("select sum(tokenStats.usageCount) as count, tokenStats.token as token from TokenStats tokenStats where tokenStats.accessTime < ?1 group by tokenStats.token")
     List<UserTokenUsage> countTokenUsageByToken(Instant before);
+
+    @Query("select sum(case when tokenStats.usageCount = 0 then 1 else tokenStats.usageCount end) as count, tokenStats.token as token, DATE_FORMAT(tokenStats.accessTime,'%Y-%m') as time, tokenStats.resource as resource " +
```
Leowisd, line 24: The number of records whose usage count equals to 0, based on month. Almost every month has records' usage count equals to 0, especially the usage counts of all records are 0 before 2020-07. Will have a look at state usage count.
image
PS, the total number of records of token stats based on month:
image

zhx828: We investigated and realized some of the POST calls were sent with zero requests.

src/main/java/org/mskcc/cbio/oncokb/querydomain/UserTokenUsageWithInfo.java
```diff
@@ -0,0 +1,14 @@
+package org.mskcc.cbio.oncokb.querydomain;
+
+import org.mskcc.cbio.oncokb.domain.Token;
+
+/**
+ * Created by Yifu Yao on 2021-1-14
+ */
+
+public interface UserTokenUsageWithInfo {
```
zhx828, line 9: An interface without actual implementation, this is new to me. Could you give me a reference how this works?

src/main/java/org/mskcc/cbio/oncokb/repository/TokenStatsRepository.java
zhx828: @Leowisd let's don't count those are 0.

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -180,6 +199,77 @@ public void updateTokenStats() {
...
+            }
+            UsageSummary currentUsageSummary = userSummary.get(userId).getSummary();
+            calculateUsageSummary(currentUsageSummary, state);
+        }
+        
+        ObjectMapper mapper = new ObjectMapper();
+        File resourceResult = new File("./resourceSummary.json");
+        File userResult = new File("./userSummary.json");
+        mapper.writeValue(resourceResult, resourceSummary);
+        mapper.writeValue(userResult, userSummary);
```
zhx828, line 242: I don't see this `mapper` has been used later.

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -180,6 +199,77 @@ public void updateTokenStats() {
...
+
+        String s3AccessKey = applicationProperties.getAws().getS3AccessKey();
+        String s3SecretKey = applicationProperties.getAws().getS3SecretKey();
+        String s3Region = applicationProperties.getAws().getS3Region();
+        AWSCredentials credentials = new BasicAWSCredentials(s3AccessKey, s3SecretKey);
+        AmazonS3 s3client = AmazonS3ClientBuilder.standard()
+                .withCredentials(new AWSStaticCredentialsProvider(credentials)).withRegion(s3Region).build();
```
zhx828, line 249: We have quite few places using s3client now. Could you write a service for this? So we could wire it using beans?

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
```diff
@@ -180,6 +199,77 @@ public void updateTokenStats() {
...
+        s3client.putObject("oncokb", RESOURCES_USAGE_SUMMARY_FILE, resourceResult);
+        s3client.putObject("oncokb", USERS_USAGE_SUMMARY_FILE, userResult);
+        resourceResult.delete();
+        userResult.delete();
+        log.info("User usage analysis completed!");
+    }
+
+    private void calculateUsageSummary(UsageSummary usageSummary, UserTokenUsageWithInfo info){
+        String resource = info.getResource().split(" ")[1];
```
zhx828, line 258: This is maybe out of this pr's scope. But I think we should have a string builder when we deal with resource. So that we have all logics together. Even better if we could model the resource.

src/main/java/org/mskcc/cbio/oncokb/web/rest/vm/usageAnalysis/UsageSummary.java
```diff
@@ -12,6 +13,11 @@
     private Map<String, JSONObject> month;
     private Map<String, Integer> year;
 
+    public UsageSummary(){
+        month = new HashMap<String, JSONObject>();
+        year = new HashMap<String, Integer>();
```
zhx828, line 18: why not move to line 13-14?

src/main/java/org/mskcc/cbio/oncokb/querydomain/UserTokenUsageWithInfo.java
Leowisd: I created it with reference to `UserTokenUsage`, which is also an interface without actual implementation.

src/main/java/org/mskcc/cbio/oncokb/web/rest/CronJobController.java
Leowisd: The method `writeValue` that can be used to serialize any Java value as JSON output, written to File provided.

zhx828: Ah yes, my bad.

zhx828 (approved):

PR #608, opened 2021-01-25
update usage-analysis-page

Leowisd:
This implemented https://github.com/oncokb/oncokb/issues/2419

And please review this PR: https://github.com/knowledgesystems/knowledgesystems-k8s-deployment/pull/105


src/main/java/org/mskcc/cbio/oncokb/web/rest/vm/usageAnalysis/ResourceModel.java
```diff
@@ -7,7 +7,7 @@
     
     public ResourceModel(String s){
         this.method =  s.split(" ")[0];
-        this.endpoint = s.split(" ")[1];
+        this.endpoint = s.split(" ")[1].replaceAll("\\%20", "");
```
zhx828, line 10: Replace the space escape to nothing?

src/main/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisController.java
```diff
@@ -127,4 +130,19 @@ private JSONObject requestData(String file)
 
         return new ResponseEntity<UsageSummary>(summary, status);
     }
+
+    @GetMapping("/usage/resource/detail")
```
zhx828, line 134: Should we just do `/user/resource/{endpoint}`?

src/main/webapp/app/config/constants.tsx
```diff
@@ -78,6 +78,8 @@ export const NOTIFICATION_TIMEOUT_MILLISECONDS = 5000;
 export const TOKEN_ABOUT_2_EXPIRE_NOTICE_IN_DAYS = 14;
 export const USAGE_TOP_USERS_LIMIT = 10000;
 export const USGAE_ALL_TIME_KEY = 'All';
+export const USGAE_ALL_TIME_VALUE = 'One year';
+export const USAGE_DETAIL_TIME_KEY = 'Detail';
```
zhx828, line 82: Should these two be updated?

src/main/webapp/app/pages/usageAnalysisPage/UserUsageDetailsTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+  defaultTimeType: number;
+};
+
+@observer
+export default class UserUsageDetailsTable extends React.Component<
+  IUserUsageDetailsTable,
+  {}
+> {
+  @observable resourcesTypeToggleValue = this.props.defaultResourcesType;
+  @observable timeTypeToggleValue = this.props.defaultTimeType;
```
zhx828, line 28: We should type this variable. I would not understand what 1 represents and what 2.

zhx828 (changes requested):

src/main/webapp/app/pages/usageAnalysisPage/UsageAnalysisPage.tsx
```diff
@@ -83,15 +124,40 @@ export default class UsageAnalysisPage extends React.Component<{
...
+    }
+
     return (
       <>
-        <Tabs defaultActiveKey="users" id="uncontrolled-tab-example">
-          <Tab eventKey="users" title="Users">
+        <Tabs
+          defaultActiveKey={this.usageType}
+          id="uncontrolled-tab-example"
+          onSelect={k => this.toggleType(UsageType[k || UsageType.USER])}
```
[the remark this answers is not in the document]
Leowisd: The type of k is `string|null`

src/main/java/org/mskcc/cbio/oncokb/web/rest/vm/usageAnalysis/ResourceModel.java
Leowisd: The target is removing unexpected space occurred in the endpoints, or sometimes endpoints will be shown like `/api/v1/annotate/%20%20%20%20%20%20%20%20%20%20%20%20%20mutations`

src/main/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisController.java
Leowisd: '/' in endpoint will cause matching error.

src/main/webapp/app/config/constants.tsx
Leowisd: I remember we think 'One Year' looks good, and the 'Detail' is not for showing

src/main/webapp/app/pages/usageAnalysisPage/UsageAnalysisPage.tsx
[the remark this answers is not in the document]
zhx828: But you have the defaultActiveKey. Even though k by definition here has the type null, but it should never be null?

src/main/webapp/app/pages/usageAnalysisPage/UsageToggleGroup.tsx
```diff
@@ -0,0 +1,30 @@
+import { observer } from 'mobx-react';
+import React from 'react';
+import { ToggleButton, ToggleButtonGroup } from 'react-bootstrap';
+import { ToggleValue } from './UsageAnalysisPage';
+
+type IUsageToggleGroupProps = {
+  defaultValue: ToggleValue;
+  toggleValueOne: ToggleValue;
+  toggleValueTwo: ToggleValue;
```
zhx828, line 9: Should this be an array?

src/main/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisController.java
zhx828: But I believe the api request has encoded the slash?

src/main/webapp/app/pages/usageAnalysisPage/UsageAnalysisPage.tsx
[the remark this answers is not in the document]
Leowisd: It should never be null, but the typescript will give error `Type 'null' cannot be used as an index type`. Any idea to solve it?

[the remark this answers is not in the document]
zhx828: You can force the type using `!`

src/main/webapp/app/pages/usageAnalysisPage/ResourceUsageDetailsTable.tsx
```diff
@@ -0,0 +1,100 @@
...
+                    ? 'Duration'
+                    : 'Time'}
+                </span>
+              ),
+              minWidth: 100,
+              accessor: 'time',
+              onFilter: (data: UsageRecord, keyword) =>
+                filterByKeyword(data.time, keyword),
+            },
+          ]}
```
zhx828, line 82: For the duplicated column definitions here and UserUsageTable, could you consolidate them?

Leowisd: I think the column definition with `id: 'resource'` in those two tables are different. And the time column also related to each `timeTypeToggleValue`. Do you think it is still necessary to consolidate them?

zhx828: Yup, I think we should considering they are both using UsageRecord model as data. They could just be defaults and you can overwrite properties at the place of calling.

Leowisd: @zhx828  Sure, have updated

zhx828 (approved):

PR #609, opened 2021-01-29
Add screenshot test

Leowisd:
This implemented https://github.com/oncokb/oncokb/issues/2318


screenshot-test/App.screenshot.js
```diff
@@ -0,0 +1,755 @@
+import puppeteer from 'puppeteer';
+import fs from 'fs';
+
+const DATA_DIR = './screenshot-test/data/';
+const CLIENT_URL = 'http://localhost:9000/';
+const SERVER_URL = 'http://localhost:9095/';
+const viewPort_1080 = {
```
zhx828, line 7: Let's change this to VIEW_PORT_1080.

screenshot-test/App.screenshot.js
```diff
@@ -0,0 +1,755 @@
...
+    browser = await puppeteer.launch(browserConfig);
+    page = await browser.newPage();
+    await page.setRequestInterception(true); // Handle UnhandledPromiseRejectionWarning: Error: Request Interception is not enabled! 
+    page.on('request', (request) => {
+      let url = request.url()
+      if (url.includes('recaptcha')){
+        request.abort();
+      }
+      else{
+        switch (url) {
```
zhx828, line 447: Can we merge this switch with the one in the Tests with login? Are they identical?

screenshot-test/App.screenshot.js
```diff
@@ -0,0 +1,755 @@
...
+            );
+            break;
+          default:
+            request.continue();
+        }
+      }     
+    });
+    await page.goto(`${CLIENT_URL}`);
+    await page.evaluate(() => {
+      localStorage.setItem('localdev', 'true');
```
zhx828, line 604: Can we move this method to an util? I can see it be used frequently in the future.

screenshot-test/App.screenshot.js
Leowisd: They are minor different, but we can merge them.

screenshot-test/App.screenshot.js
Leowisd: Actually it seems cannot be moved to an util, or the `localStorage` will ne undefined.

screenshot-test/App.screenshot.js
```diff
@@ -0,0 +1,596 @@
...
+  today.setDate(today.getDate()+100);
+  current.forEach(token => {
+    token.expiration = today.toISOString();
+  })
+  return current;
+}
+let apiAccountToken = fs.readFileSync(`${DATA_DIR}api-account-token.json`).toString();
+apiAccountToken = JSON.stringify(updateTokenExpirationDate(JSON.parse(apiAccountToken)));
+let userToken = fs.readFileSync(`${DATA_DIR}api-users-tokens.json`).toString();
+userToken = JSON.stringify(updateTokenExpirationDate(JSON.parse(userToken)));
```
Leowisd, line 55: Add a process to set expiration date dynamically, or the screenshots of Account Settings Page and User Details Page will be different daily(because of to expiration date).

screenshot-test/App.screenshot.js
```diff
@@ -0,0 +1,596 @@
...
+    path: `${LATEST_SNAPSHOTS_DIR}${name}-snap.png`,
+    fullPage: true
+  }
+}
+
+if (!fs.existsSync(LATEST_SNAPSHOTS_DIR)){
+  fs.mkdirSync(LATEST_SNAPSHOTS_DIR);
+}
+
+function getRes(url){
```
zhx828, line 69: Let's make these naming more clear.
```suggestion
function getMockResponse(url){
```

screenshot-test/App.screenshot.js
```diff
@@ -0,0 +1,596 @@
...
+    });
+  })
+
+  it('Home Page', async() => {
+    await page.goto(`${CLIENT_URL}`);
+    await page.setViewport(VIEW_PORT_1080);
+    await page.waitFor(WAITING_TIME);
+    let image = await page.screenshot(getScreenshotConfig('Home Page without Login'));
+    expect(image).toMatchImageSnapshot({ customSnapshotIdentifier: 'Home Page without Login' });
+  })
```
zhx828, line 455: I just noticed that these tests designed for with/out login are identical except some string which can be changed through input. I believe you want to design test for both with and without login, correct? Maybe consolidate these tests to a method?

jest-puppeteer.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+    launch: {
+        dumpio: true,
+        headless: true
+    },
+    browser: 'chromium',
```
zhx828, line 6: Can we run on ie/safari as well?

zhx828 (changes requested):

screenshot-test/App.screenshot.js
Leowisd: I've considered it. But there would be some async errors if wrap these in a method. Maybe we could use it now and I will do more research later.

jest-puppeteer.config.js
Leowisd: Currently the puppeteer only supports Chrome and Firefox, but I think it will supports more browsers in the future. However, it currently can emulate different devices with different userAgents.

PR #671, opened 2021-06-22
Generate unapproved user tables in the email based on the slack discussions

benzuzu:
### Unapproved users are divided into two groups based on license type and sent through a weekly email

This fixes https://github.com/oncokb/oncokb/issues/2614

- Changed lists to be based on license type
- Added timestamps for creation date and last discussed date
- Removed other columns for better spacing (email, job title)

Screen Shot 2021-07-06 at 12 10 52 PM


src/main/resources/templates/mail/listOfUnapprovedUsers.html
```diff
@@ -23,17 +24,52 @@
     <th>Job Title</th>
     <th>Company</th>
     <th>Country</th>
+    <th>View Request in Slack</th>
   </tr>
   </thead>
   <tbody>
-  <tr th:each="user: ${users}">
+  <tr th:each="user: ${undiscussedUsers}">
```
zhx828, line 31: Haven't used before, but we can consider create a fragment for the table so it could be reused. https://www.baeldung.com/spring-thymeleaf-fragments

zhx828 (changes requested):
Like we discussed, the following points needs to be updated
- We no longer need to save the message id in the database 
- The blocks need to have an id to be parsed
- Try to reuse the html fragments

src/main/resources/config/application-dev.yml
```diff
@@ -37,7 +37,7 @@ spring:
       indent-output: true
   datasource:
     type: com.zaxxer.hikari.HikariDataSource
-    url: jdbc:mysql://localhost:3306/oncokb-public?useUnicode=true&characterEncoding=utf8&useSSL=false&useLegacyDatetimeCode=false&serverTimezone=UTC&createDatabaseIfNotExist=true
+    url: jdbc:mysql://localhost:3306/oncokb-public?useUnicode=true&characterEncoding=utf8&useSSL=false&useLegacyDatetimeCode=false&serverTimezone=UTC&createDatabaseIfNotExist=true&allowPublicKeyRetrieval=true
```
zhx828, line 40: Why are we adding `allowPublicKeyRetrieval`?

src/main/resources/templates/fragments/tables.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
+<html xmlns:th="http://www.thymeleaf.org" lang="en">
+<head>
+  <title>Table Fragments</title>
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+  <link rel="shortcut icon" th:href="@{|${baseUrl}/favicon.ico|}" />
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/css/bootstrap.min.css" integrity="sha384-JcKb8q3iqJ61gNV9KGb8thSsNjpSL0n8PARn9HuZOnIxN0hoP+VmmDGMN5t9UJ0Z" crossorigin="anonymous">
```
zhx828, line 7: If we don't use the style. We should not include the stylesheet.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -501,8 +555,20 @@ private ButtonElement buildButton(String text, String value, ActionId actionId,
...
 
-    private LayoutBlock buildPlainTextBlock(String text) {
-        return SectionBlock.builder().text(PlainTextObject.builder().text(text).build()).build();
+    private LayoutBlock buildPlainTextBlock(String text, String id) {
+        return SectionBlock.builder().text(PlainTextObject.builder().text(text).build()).blockId(id).build();
+    }
+
+    private Optional<LayoutBlock> getSectionBlockWithId(List<LayoutBlock> blocks, String id) {
+        for (LayoutBlock block : blocks) {
+            if(block.getClass().getName().equals("com.slack.api.model.block.SectionBlock")) {
```
zhx828, line 564: What else could it be?

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
benzuzu: There are also dividerblocks that cannot be parsed into sectionblocks, and regular layout blocks don't have block id's so they must be parsed.

src/main/resources/config/application-dev.yml
benzuzu: For some reason my database would not work for some time- it kept throwing an error "public key retrieval is not allowed." I couldn't even run the application so I went online and found that adding this query worked. I can remove it while I fix the other issues.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -404,6 +421,43 @@ private boolean withClarificationNote(UserDTO userDTO) {
...
+        Slack slack = Slack.getInstance();
+        try {
+            String daysAgoTs = Long.toString(Instant.now().getEpochSecond() - DAY_IN_SECONDS * daysAgo);
+            ConversationsHistoryResponse conversationsHistory = slack.methods().conversationsHistory(r -> r
+                .token(applicationProperties.getSlack().getSlackBotOAuthToken())
+                .channel(applicationProperties.getSlack().getUserRegistrationChannelID())
+                .oldest(daysAgoTs)
+                .inclusive(true));
+            if(conversationsHistory.isHasMore()) {
+                log.warn("Number of approval requests has exceeded maximum limit. Some requests will not be displayed in weekly email!");
```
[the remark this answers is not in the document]
benzuzu: I believe the conversationsHistory has a limit of 99 or 100 messages. I didn't see any way around it but I didn't look very long so I can check again.

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
```diff
@@ -311,18 +312,33 @@ public void sendMailToUserWhenTokenExposed(UserDTO user, ExposedToken token) {
...
+
+        for(UserMessagePair userMessagePair : users) {
+            addUserMailsRecord(userMessagePair.getUserDTO(), LIST_OF_UNAPPROVED_USERS, applicationProperties.getEmailAddresses().getTechDevAddress(), "bot");
+        }
+
+        List<UserMessagePair> discussedUsers = users.stream().filter(user -> Objects.nonNull(user.getMessage().getReplyCount())).collect(Collectors.toList());
+        for(UserMessagePair user : discussedUsers) {
+            users.remove(user);
+            users.add(user);
+        }
```
zhx828, line 328: What's going on here?

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
```diff
@@ -311,18 +312,33 @@ public void sendMailToUserWhenTokenExposed(UserDTO user, ExposedToken token) {
...
     @Async
-    public void sendUnapprovedUsersEmail(int daysAgo, List<UserDTO> users){
+    public void sendUnapprovedUsersEmail(int daysAgo, List<UserMessagePair> users){
         if (users == null || users.isEmpty()) {
             return;
         }
+
+        for(UserMessagePair userMessagePair : users) {
+            addUserMailsRecord(userMessagePair.getUserDTO(), LIST_OF_UNAPPROVED_USERS, applicationProperties.getEmailAddresses().getTechDevAddress(), "bot");
+        }
```
zhx828, line 322: Let's remove this for now. I don't think it's useful adding a record  for this.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -228,6 +235,13 @@ private LayoutBlock buildChannelMentionBlock() {
         return blocks;
     }
 
+    private List<LayoutBlock> buildUserIdBlocks(UserDTO userDTO) {
+        List<LayoutBlock> blocks = new ArrayList<>();
+        blocks.add(SectionBlock.builder().text(PlainTextObject.builder().text("User ID: " + userDTO.getId()).build()).blockId(USER_ID.getId()).build());
```
zhx828, line 240: Through some initial digging, I don't think we can change the font size/color. (I might be wrong though)
If this is the case, should we use a https://api.slack.com/reference/block-kit/blocks#context block? It looks less dominated in terms of styles.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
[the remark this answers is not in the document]
zhx828: Could you add a warning in the email when this happen instead of only prints in the log? We do not want to miss any messages.

src/main/resources/config/application-dev.yml
```diff
@@ -143,13 +143,17 @@ application:
...
   licensed-domains:
   trialed-domains:
   account-approval-whitelist:
   token-usage-check:
   token-usage-check-whitelist:
+  slack:
+    user-registration-webhook:
+    user-registration-channel-id:
+    slack-bot-oauth-token:
+    slack-base-url:
```
zhx828, line 156: Could you add some comments here explaining what they are for?

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
benzuzu: Good idea, this stands out much less.

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
benzuzu: This sorts the list by whether or not a user has been discussed. So all of the users who have been discussed get moved to the bottom. I wasn't able to iterate through the initial list while changing the size of the list so I thought a solution would be to make a new list and iterate through that and move each corresponding element to the end.

zhx828: Should this just be a sort of the `users`? https://stackoverflow.com/a/34646172 @benzuzu

benzuzu: Ah, yeah this works better. Changed and should be good now.

zhx828 (approved):
Great work! 💯

PR #676, opened 2021-07-14
Add option to send denial email citing embargo policy

benzuzu:
### Embargo Country Text Match will Trigger Warning in Slack Integration

If a user registers through an organization located in an embargoed country (list specified in application properties), a warning will appear in the Slack request:

Screen Shot 2021-07-14 at 11 49 05 AM

### Embargo Email

There is now an option in Slack to deny a user due to embargo policy. The following email will be sent to the user:

Screen Shot 2021-07-20 at 12 12 24 PM

### Changes to Slack Integration

The Slack integration has been adjusted accordingly.

- Non-approve buttons have been moved to a dropdown menu for organization
- Clarification and Denial emails can be sent in this menu

Screen Shot 2021-07-14 at 11 41 37 AM
Screen Shot 2021-07-14 at 11 41 48 AM


src/main/java/org/mskcc/cbio/oncokb/domain/UserStatusChecks.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class UserStatusChecks {
+    UserDTO userDTO;
+
+    boolean trialAccountActivated;
+    boolean trialAccountInitiated;
+    boolean academicClarificationEmailSent;
+    boolean embargoEmailSent;
+
+    public UserStatusChecks(UserDTO userDTO, UserService userService, SlackService slackService) {
```
zhx828, line 18: We should not pass the service layout objects directly but the actual value to initialize the UserStatusChecks

src/main/java/org/mskcc/cbio/oncokb/config/application/ApplicationProperties.java
```diff
@@ -171,6 +176,10 @@ public void setFrontend(FrontendProperties frontend) {
         return getDomains(this.getTrialedDomains());
     }
 
+    public List<String> getEmbargoedCountriesList() {
+        return getDomains(this.getEmbargoedCountries());
+    }
+
     private List<String> getDomains(String domainStr) {
```
zhx828, line 183: Maybe we should rename this method now  to `stringSplitAndTrim`? And good name recommendation?

zhx828 (changes requested):
Looks good, just some minor tweaks. Let's also discuss the issue on next Tuesday meeting.

src/main/java/org/mskcc/cbio/oncokb/config/application/ApplicationProperties.java
benzuzu: How does `toList` sound? Or `stringToList`, `getList`

benzuzu:
Closed until further notice

PR #677, opened 2021-07-15
Minor fixes to weekly email and slack integration

benzuzu:
- Empty fields no longer show up as `**`
- If a list is empty the table will not show up in weekly email

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
@@ -137,8 +137,12 @@ public class SlackService {
 
     private TextObject getTextObject(String title, String content) {
         StringBuilder sb = new StringBuilder();
-        sb.append(title + ":\n");
-        sb.append("*" + content + "*");
+        if (title != null) {
+            sb.append(title + ":\n");
+        }
+        if (content != null) {
+            sb.append("*" + content + "*");
+        }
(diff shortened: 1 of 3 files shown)


src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -137,8 +137,12 @@ public ActionId getActionId(BlockActionPayload blockActionPayload) {
 
     private TextObject getTextObject(String title, String content) {
         StringBuilder sb = new StringBuilder();
-        sb.append(title + ":\n");
-        sb.append("*" + content + "*");
+        if (title != null) {
```
zhx828, line 140: Maybe do a StringUtils.isNotEmpty check instead? If the content is empty string, you also don't want to include.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -137,8 +137,12 @@ public ActionId getActionId(BlockActionPayload blockActionPayload) {
...
+            sb.append(title + ":\n");
+        }
+        if (content != null) {
+            sb.append("*" + content + "*");
```
zhx828, line 144: Specially here, when the content is empty, you will still print two starts?

zhx828 (changes requested):

zhx828 (approved):

PR #683, opened 2021-07-27
Fixed clarification email bug

benzuzu:
Slack blocks were not collapsing automatically when the academic clarification email was sent. Now they automatically collapse to the following block:

Screen Shot 2021-07-27 at 11 34 38 AM


zhx828 (approved):
Looks good. Thanks for the quick fix.

PR #684, opened 2021-07-30
Add status column

benzuzu:
- user_details table now has an account status column indicating more specific statuses
- API call to populate status for past users


benzuzu:
Push to new branch

PR #708, opened 2021-09-27
Update jdl file

zhx828:


jhipster-jdl.jh
```diff
@@ -21,19 +21,19 @@ entity UserDetails {
...
     country String,
     address String,
-    additionalInfo String
+    additionalInfo TextBlob
 }
 
 entity UserMails {
 	sentDate Instant required,
     sentBy String required,
-    mailType MailType,
```
zhx828, line 30: Let's get rid of the enum restriction on the mail Type here since the list keeps expending quite frequently

jhipster-jdl.jh
```diff
@@ -21,19 +21,19 @@ entity UserDetails {
     city String,
     country String,
     address String,
-    additionalInfo String
+    additionalInfo TextBlob
```
zhx828, line 24: String by default uses 255 chars. But we need a bigger size here.

jhipster-jdl.jh
calvinlu3: Yeah that is good point. MailType will need more values for when we add company entity.

calvinlu3 (approved):

PR #709, opened 2021-09-27
Regenerate codebase based on the latest JDL file

zhx828:


.jhipster/UserDetails.json
```diff
@@ -11,7 +11,7 @@
       "fieldType": "String"
     },
     {
-      "fieldName": "company",
+      "fieldName": "companyName",
```
zhx828, line 14: We need to change the company String filed to companyName so it would not conflict with the company table,

jhipster-jdl.jdl
```diff
@@ -66,35 +66,23 @@ relationship OneToOne {
...
 	UserMails{user} to User
-}
-
-relationship ManyToOne {
 	CompanyDomain{company} to Company
-}
-
-relationship ManyToOne {
 	UserDetails{company} to Company
 }
```
zhx828, line 73: We could put everything together

PR #712, opened 2021-09-28
Make Redis caching optional

calvinlu3:
Fixes [oncokb/#2688](https://github.com/oncokb/oncokb/issues/2688#issue-999591831)

Changes

- Use RedisProperties and RedisType file from oncokb-meta, which introduces a new _enabled_ property
     - Redis is enabled by default for the _prod_ profile
- When Redis is enabled, the configurations for cache manager/cache are unchanged. When Redis is disabled, then to avoid the "No bean of type org.springframework.cache.CacheManager" error, we can create a custom cache manager that controls in-memory caches. 
     - The other option is to skip CacheConfiguration when Redis is disabled and make CacheManager optional using Java8's Optional class.


src/main/java/org/mskcc/cbio/oncokb/config/CacheConfiguration.java
```diff
@@ -29,7 +31,18 @@
...
 @EnableCaching
 public class CacheConfiguration {
+
+    private final String PROP_PREFIX = "application.redis";
+    private final String PROP_NAME = "enabled";
+
+    @Bean
+    @ConditionalOnProperty(prefix = PROP_PREFIX, name = PROP_NAME, havingValue = "false")
+    public org.springframework.cache.CacheManager cacheManager(ApplicationProperties applicationProperties){
+        return new CustomCacheManager();
```
calvinlu3, line 41: Since the custom cache manager doesn't really do anything special in the getCache method, we can probably use a cache manager from spring instead (like ConcurrentMapCacheManager).

src/main/java/org/mskcc/cbio/oncokb/config/application/ApplicationProperties.java
```diff
@@ -3,6 +3,7 @@
 import org.apache.commons.lang3.StringUtils;
 import org.mskcc.cbio.oncokb.domain.enumeration.ProjectProfile;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.mskcc.oncokb.meta.model.application.RedisProperties;
```
zhx828, line 6: I guess this reference is no longer needed?

zhx828 (approved):
Looks good, only one minor comment.

zhx828:
Nicely done! 💯

src/main/java/org/mskcc/cbio/oncokb/config/cache/CustomCacheManager.java
```diff
@@ -0,0 +1,46 @@
...
+	 * does not exist or could be not created
+	 */
+    @Override
+    public Cache getCache(String name) {
+        Cache cache = this.cacheMap.get(name);
+        if (cache == null){
+            cache = new ConcurrentMapCache(name, new ConcurrentHashMap<>(), true);
+            cacheMap.put(name, cache);
+        }
+        return cache;
```
zhx828, line 43: Oh I just realize that we actually store the content when the redis is disabled. This memory cache will continue building up?

calvinlu3: Good point, the cache will buildup and we might run out of memory. I can evict all entries from the cache if cache != null before returning the cache.

src/main/java/org/mskcc/cbio/oncokb/config/application/ApplicationProperties.java
calvinlu3: Its the type for the **redis** field.

src/main/java/org/mskcc/cbio/oncokb/config/CacheConfiguration.java
```diff
@@ -36,8 +36,8 @@
     private final String PROP_NAME = "enabled";
 
     @Bean
-    @ConditionalOnProperty(prefix = PROP_PREFIX, name = PROP_NAME, havingValue = "false")
-    public org.springframework.cache.CacheManager cacheManager(ApplicationProperties applicationProperties){
+    @ConditionalOnProperty(prefix = PROP_PREFIX, name = PROP_NAME, havingValue = "false", matchIfMissing = true)
```
calvinlu3, line 39: Conditions are evaluated before ConfigurationProperties properties are available in Spring, so when "application.redis.enabled" is not present in yml file, then the custom cache manager is not created. matchIfMissing sets the default behavior when that property is missing.

PR #720, opened 2021-10-03
Rework token_stats storage

benzuzu:
### The token_stats table is now completely cleared periodically

In order to prevent the table from overflowing, all entries are deleted when the updateTokenStats() cronjob is called. This cronjob now just adds usage to each token's current usage instead of replacing it.

### Old token_stats backups are stored in a new directory liquibase/old-token-stats as .csv files

When TokenStats are saved, they are now also appended to a .csv file in liquibase/data. When updateTokenStats() is called, the file is moved to old-token-stats and the name is changed to include the current timestamp. A new file then takes its place.

This fixes https://github.com/oncokb/oncokb/issues/2648

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/domain/TokenStats.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/domain/TokenStats.java
@@ -144,4 +144,13 @@ public class TokenStats implements Serializable {
             ", usageCount=" + getUsageCount() +
             "}";
     }
+
+    public String toCSV() {
+        return getId() +
+            ";" + getAccessIp() +
+            ";" + getResource() +
+            ";" + getAccessTime() +
+            ";" + getUsageCount() +
+            ";" + getToken().getId();
+    }
 }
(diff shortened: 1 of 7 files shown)


zhx828 (changes requested):
Thanks for working on this, here are few concerns about this pr
- [ ] We should avoid editing files locally because in production the instance is scalable. You might end up editing different files in different instances. Use s3 instead.
- [ ] Writing on s3 file on every token save seems too expensive(file operation and network latency). But I might as well be wrong. We need to explore more on this topic. S3 seems to have very good performance per ad https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html
- [ ] We could consider writing to the database and wrap the data on daily basis to decrease the # of connections
- [ ] When fetching token usage, you should add a timestamp instead of using Instant now. Getting all tokens takes time. There might be more token usages created after fetching before deleting. Then when you call `clearTokenStats`, you might delete additional tokens. If you use a timestamp, then only delete the token usage before that timestamp should avoid the issue. We could use UTC+0  for simplicity.

PR #733, opened 2021-10-12
Extend the wait time for Dx/Px tabs in the LoE page

zhx828:
Experiment: assume the transition takes longer than 2s.


zhx828:
This does not seem to help with the situation.

PR #737, opened 2021-10-13
Consolidate tabs in the gene page/annotation pages together

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/2687


zhx828:
@calvinlu3 could you just do some spot checks, just in case I missed anything?

PR #742, opened 2021-10-19
Allow ChangedAnnotationTable changes header based on the usage

calvinlu3:
Added new column headers for when the drug is changed rather than the level.

### DIFF
--- a/src/main/webapp/app/pages/newsPage/ChangedAnnotatonListItem.tsx
+++ b/src/main/webapp/app/pages/newsPage/ChangedAnnotatonListItem.tsx
@@ -1,4 +1,7 @@
-import { CHANGED_ANNOTATION_COLUMNS } from 'app/pages/newsPage/NewsPageContent';
+import {
+  CHANGED_ANNOTATION_LEVEL_COLUMNS,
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/pages/newsPage/ChangedAnnotatonListItem.tsx
```diff
@@ -1,26 +1,46 @@
...
   let longestRow = 0;
   if (props.data.length > 0) {
     longestRow = _.sortBy(props.data, row => -row.content.length)[0].content
       .length;
   }
+
+  let annotationColumnHeader = undefined;
+  switch (props.columnHeaderType) {
+    case AnnotationColumnHeaderType.DRUG:
+      annotationColumnHeader = CHANGED_ANNOTATION_DRUG_COLUMNS;
```
zhx828, line 31: You need a break statement here. I'm surprised it still works in the drug table?

calvinlu3: Good catch, I missed this when setting the default to the level headers

PR #744, opened 2021-10-22
View and Update Company Pages

calvinlu3:
This is related to:  [issue #2695](https://github.com/oncokb/oncokb/issues/2695)
Depends on [company logic PR](https://github.com/oncokb/oncokb-public/pull/730)

#### In Progress
 - [x] After company creation, link to update company page to let admin add all users or selected users to company
 - [x] Logic for unlinking users 

#### Progress from review
 - [x] On the frontend, make sure that certain license status changes are not allowed (ie. Regular -> Trial Expired)
 - [x] Instead of using a toaster message when domain being added is invalid format, show error message under text input

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/web/rest/UserResource.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/web/rest/UserResource.java
@@ -7,6 +7,8 @@ import org.apache.http.protocol.HTTP;
 import org.mskcc.cbio.oncokb.config.Constants;
 import org.mskcc.cbio.oncokb.domain.Token;
 import org.mskcc.cbio.oncokb.domain.User;
+import org.mskcc.cbio.oncokb.domain.UserDetails;
+import org.mskcc.cbio.oncokb.repository.UserDetailsRepository;
 import org.mskcc.cbio.oncokb.repository.UserRepository;
 import org.mskcc.cbio.oncokb.security.AuthoritiesConstants;
 import org.mskcc.cbio.oncokb.service.MailService;
@@ -84,13 +86,22 @@ public class UserResource {
 
     private final UserRepository userRepository;
 
+    private final UserDetailsRepository userDetailsRepository;
+
     private final MailService mailService;
 
     private final TokenService tokenService;
 
-    public UserResource(UserService userService, UserRepository userRepository, MailService mailService, TokenService tokenService) {
+    public UserResource(
+        UserService userService, 
+        UserRepository userRepository,
+        UserDetailsRepository userDetailsRepository,
+        MailService mailService, 
+        TokenService tokenService
+    ) {
         this.userService = userService;
         this.userRepository = userRepository;
+        this.userDetailsRepository = userDetailsRepository;
         this.mailService = mailService;
         this.tokenService = tokenService;
     }
@@ -150,7 +161,8 @@ public class UserResource {
     @PreAuthorize("hasAuthority(\"" + AuthoritiesConstants.ADMIN + "\")")
     public ResponseEntity<UserDTO> updateUser(
         @Valid @RequestBody UserDTO userDTO,
-        @NotNull @RequestParam Boolean sendEmail
+        @NotNull @RequestParam Boolean sendEmail,
+        @NotNull @RequestParam Boolean unlinkUser
     ) {
         log.debug("REST request to update User : {}", userDTO);
         Optional<User> existingUser = userRepository.findOneByEmailIgnoreCase(userDTO.getEmail());
@@ -161,6 +173,16 @@ public class UserResource {
         if (existingUser.isPresent() && (!existingUser.get().getId().equals(userDTO.getId()))) {
             throw new LoginAlreadyUsedException();
         }
+
+        // Remove mapping between the user and the company
+        if(existingUser.isPresent() && unlinkUser){
+            Optional<UserDetails> userDetails = userDetailsRepository.findOneByUser(existingUser.get());
+            userDetails.ifPresent(ud -> {
+                ud.setCompany(null);
+                userDetailsRepository.save(ud);
+            });
+        }
+
         Optional<UserDTO> updatedUser = userService.updateUser(userDTO);
 
         if(updatedUser.isPresent() && sendEmail && updatedUser.get().isActivated()) {
@@ -263,4 +285,20 @@ public class UserResource {
         }
         return new ResponseEntity<>(tokenService.findByUser(optionalUser.get()), HttpStatus.OK);
     }
+
+    @PostMapping("/users/tokens")
+    public ResponseEntity<List<Token>> getUsersTokens(@RequestBody List<String> logins){
+        log.debug("REST request to get tokens of each user");
+        List<Token> usersTokens = new ArrayList<>();
+        for(String login: logins){
+            Optional<User> optionalUser = userService.getUserWithAuthoritiesByLogin(login);
+            if (optionalUser.isPresent()) {
+                usersTokens.addAll(tokenService.findByUser(optionalUser.get()));
+            }else{
+                return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+            }
+        }
+        return new ResponseEntity<>(usersTokens, HttpStatus.OK);
+    }
+
 }
--- a/src/main/webapp/app/pages/companyManagement/CompanyDetailsPage.tsx
+++ b/src/main/webapp/app/pages/companyManagement/CompanyDetailsPage.tsx
@@ -1,0 +1,184 @@
+import _ from 'lodash';
+import React from 'react';
+import { Col, Row } from 'react-bootstrap';
+import { observer } from 'mobx-react';
+import client from 'app/shared/api/clientInstance';
+import { remoteData } from 'cbioportal-frontend-commons';
+import styles from './CompanyDetailsPage.module.scss';
+import { CompanyDTO } from 'app/shared/api/generated/API';
+import OncoKBTable, {
+  SearchColumn,
+} from 'app/components/oncokbTable/OncoKBTable';
+import { getSectionClassName } from 'app/pages/account/AccountUtils';
+import { notifyError } from 'app/shared/utils/NotificationUtils';
+import { filterByKeyword } from 'app/shared/utils/Utils';
+import { PAGE_ROUTE } from 'app/config/constants';
+import LoadingIndicator, {
+  LoaderSize,
+} from '../../components/loadingIndicator/LoadingIndicator';
+import { Link } from 'react-router-dom';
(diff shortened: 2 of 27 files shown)


zhx828 (changes requested):
Sorry about the long list, trying to put everything I found together.

### Create company page
- [x] The license status should only be Regular/Trial. Other statuses should be available in the update page.
- [x] After adding the domain in the input box, if I click enter key, it should be automatically added if the domain format is valid.
- [x] After creating, we still need a successs message showing on the top.

### Update company page
- [x] let's remove the Add New Company from the Quick Tools
- [x] `Add All Related Users` button should be on the same line of the user dropdown. It was a bit confusing using it.Please see the following for the mockup
Screen Shot 2021-10-29 at 4 21 07 PM

- [x] In the user table, do not show city, country. Highlight license type if it's not the same as to company's
- [x] The board color of the Add button for the Company Domains should be light grey, similar to others. We have a $oncokb-border-color variable you can use.
- [ ] After adding the user to the company, the user license is not updated.
- [ ] Once admin changes company license, we should update all associated users' license.

####  License Status change
- [x] License Status dropdown should show with the following order, Regular, Trial, Trial Expired, Expired, Unknown
- [x] Let's hide the unknown status for now to avoid dealing with it.
- [ ] License status change should show a confirm dialog with old/new changes and potential side effects. Side effects will be different for different scenarios. The key here is to give warnings about what will happen after making this change.
- [ ] Regular-Trial, all users should be turned to trialed.
- [ ] Regular->Trial Expired, this should not be allowed.
- [ ] Regular->Expired, all accounts will be terminated right away.
- [ ] Trial->Regular, all users should be turned to regular.
- [ ] Trial->Trial Expired, all accounts will be terminated right away.
- [ ] Trial->Expired, all accounts will be terminated right away.
- [ ] If company has trial status, after adding the user, user should receive the trial activation link automatically (to be discussed)

zhx828 (changes requested):
It looks good, some minor comments below. Once you finish these, I will start to do the code review.
- [x] Do not allow admin to change the user's license if they are associated with the company. In current implementation, the user's license type is always come from the company. Updating the license in the frontend does not work anymore)
- [x] Update document title to user name  in user page, to company name in company page. See component `DocumentTitle`
- [x] Let's don't disable user when removed from company association list
- [x] Add a tooltip on the `Add All Related Users` to explain the purpose of this button
- [ ] In the company users list, can we show whether user is on trial?

calvinlu3:
#### Update
Pertaining to the first bullet point. The license type buttons are disabled when the user is associated with a company. The admin might be confused why buttons are disabled, so an InfoIcon is included explaining. 
example

For the last bullet point, a user is on trial if they have a token that is non-renewable. Currently, I send a GET request for each company user, but there may be a more efficient way to do this.

src/main/webapp/app/shared/list/FormListField.tsx
```diff
@@ -0,0 +1,159 @@
...
+  onDelete: (item: string) => void;
+};
+
+@observer
+class ListBox extends React.Component<IListBoxProps> {
+  render() {
+    return (
+      <div
+        style={{
+          border: '1px solid #ced4da',
```
zhx828, line 20: Let's use the scss variable?

src/main/webapp/app/shared/list/FormListField.tsx
```diff
@@ -0,0 +1,159 @@
...
+        }}
+      >
+        <div className={styles.main}>
+          {this.props.list.map(item => (
+            <div key={item} className={styles.content_container}>
+              <span
+                className={styles.delete}
+                onClick={() => this.props.onDelete(item)}
+              >
+                <i className="fa fa-times-circle"></i>
```
zhx828, line 36: Let's move this to a functional component similar to InfoIcon.

src/main/webapp/app/shared/list/FormListField.tsx
```diff
@@ -0,0 +1,159 @@
...
+};
+
+@observer
+export class FormListField extends React.Component<IFormListFieldProps> {
+  @observable isInputValid = false;
+  @observable inputText = '';
+  @observable errorMessage = '';
+
+  @autobind
+  listLengthValidation(value: any, context: any, input: any, cb: any) {
```
zhx828, line 63: I think we can type the `cb`?

src/main/webapp/app/shared/list/FormListField.tsx
```diff
@@ -0,0 +1,159 @@
...
+          <Col>
+            <label
+              className={this.props.boldLabeL ? 'font-weight-bold' : ''}
+              htmlFor="listItemInput"
+            >
+              {this.props.labelText}
+            </label>
+            <div className="input-group">
+              <AvInput
+                id="listItemInput"
```
zhx828, line 110: I'm sure the id will be duplicated if multiple FormListField is required. If we cannot avoid using the id, maybe we should require a `id` attr on the component?

src/main/webapp/app/shared/list/FormListField.tsx
```diff
@@ -0,0 +1,159 @@
...
+                }}
+                validate={{
+                  customLength: this.listLengthValidation,
+                }}
+                onKeyPress={this.onEnterKey}
+              />
+              <div className="input-group-append">
+                <Button
+                  className={`${styles.add} btn btn-light`}
+                  style={{ zIndex: 0 }}
```
zhx828, line 126: Why do we need to specify the zindex?

src/main/webapp/app/shared/list/FormListField.tsx
```diff
@@ -0,0 +1,159 @@
...
+                >
+                  Add
+                </Button>
+              </div>
+              <span
+                style={{
+                  width: '100%',
+                  marginTop: '0.25rem',
+                  fontSize: '80%',
+                  color: '#dc3545',
```
zhx828, line 142: Let's use class name `text-danger`

src/main/webapp/app/pages/companyManagement/CompanyDetailsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+import LoadingIndicator, {
+  LoaderSize,
+} from '../../components/loadingIndicator/LoadingIndicator';
+import { Link } from 'react-router-dom';
+import { QuickToolButton } from '../userPage/QuickToolButton';
+
+@inject('routing')
+@observer
+export default class CompanyDetailsPage extends React.Component<{
+  routing: RouterStore;
```
zhx828, line 27: I don't see the usage of `routing`

src/main/webapp/app/pages/companyManagement/CompanyDetailsPage.tsx
```diff
@@ -0,0 +1,203 @@
...
+  @observable companies: CompanyDTO[] = [];
+  @observable areCompaniesLoaded = false;
+
+  constructor(props: Readonly<{ routing: RouterStore; match: match }>) {
+    super(props);
+    this.getCompanies();
+  }
+
+  @action
+  getCompanies() {
```
zhx828, line 39: If you want to use the status of loading companies, maybe use `remoteData` instead?

src/main/webapp/app/pages/companyPage/CompanyPage.tsx
```diff
@@ -0,0 +1,521 @@
...
+import Select from 'react-select';
+import DocumentTitle from 'react-document-title';
+import { DefaultTooltip } from 'cbioportal-frontend-commons';
+
+interface MatchParams {
+  id: string;
+}
+
+interface ICompanyPageProps extends RouteComponentProps<MatchParams> {
+  windowStore: WindowStore;
```
zhx828, line 43: No usage of windowStore?

src/main/webapp/app/pages/companyPage/CompanyPage.tsx
```diff
@@ -0,0 +1,521 @@
...
+                              'Include at least one domain. ie) oncokb.org'
+                            }
+                            boldLabeL
+                          />
+                          <div className="form-group">
+                            <div className={'font-weight-bold'}>
+                              Add Users to Company
+                            </div>
+                            <div
+                              style={{ display: 'flex', flexDirection: 'row' }}
```
zhx828, line 419: row is the default flex direction.

src/main/webapp/app/pages/companyPage/CompanyPage.tsx
```diff
@@ -0,0 +1,521 @@
...
+        label: user.email,
+        value: user.id,
+        user,
+      }));
+  }
+
+  @action
+  async getAllUsersTokens() {
+    for (const user of this.companyUsers) {
+      const tokens = await client.getUserTokensUsingGET({
```
zhx828, line 141: This is ok for now, but might below the performance when too many users are associated with the company. Let's leave it as it but could you add a comment here?

zhx828 (changes requested):
Looks good! Just some minor comments. Great work 👍🏼

src/main/webapp/app/shared/list/FormListField.tsx
calvinlu3: When there is a react-select Select component, the dropdown appears under the button.
image

src/main/webapp/app/pages/companyPage/CompanyPage.tsx
calvinlu3: From weekly meeting: Create another endpoint that will take a list of users and return a list of users tokens

src/main/webapp/app/shared/list/FormListField.tsx
calvinlu3: Yeah, the callback is a function that takes a boolean or string and returns void.

src/main/webapp/app/shared/list/FormListField.tsx
calvinlu3: Good point. I used the `id` attribute on `AvInput`, so that the label is bound to it.

src/main/java/org/mskcc/cbio/oncokb/web/rest/UserResource.java
```diff
@@ -263,4 +285,20 @@ private boolean onlyContainsAllowedProperties(Pageable pageable) {
         }
         return new ResponseEntity<>(tokenService.findByUser(optionalUser.get()), HttpStatus.OK);
     }
+
+    @PostMapping("/users/tokens")
+    public ResponseEntity<List<List<Token>>> getUsersTokens(@RequestBody List<String> logins){
```
zhx828, line 290: I think we should just return List<Token>, an 2d array is not super useful and complicate the maintenance.

zhx828 (approved):

PR #750, opened 2021-11-03
Support exclusion rule in the academic-email-clarify-domain

zhx828:
This allows us to exclude a specific domain from a broader list. For instance we could use `.com,!test.com`
This fixes https://github.com/oncokb/oncokb/issues/2731


calvinlu3 (approved):
Looks good.

PR #762, opened 2021-11-22
Mixed Company Usage

calvinlu3:
This fixes [issue 2719](https://github.com/oncokb/oncokb/issues/2719).
Also addresses [remove whitelist ticket](https://github.com/oncokb/oncokb/issues/2786) and [remove licensed/trialed domains](https://github.com/oncokb/oncokb/issues/2683).

## Changes
- Update Company and Company Domain relationship to many to many.
- Fixed some issues related to the `add company dropdown` that popped up from testing. ( For non-admins, we were still fetching from the /companies endpoint, and license model update wasn't shown properly).
- Updated logic for activating a user after registration and verified their email
     - When email domain matches a MICRO tiered company --> manual approval (send user registration to slack)
     - When email domain matches a REGULAR tiered company
          - If the license type the user registers with aligns with company, then automatically approve
          - Otherwise, align with company license, approve, and then send email with clarification.
    - When email domain matches several companies --> find a company that is on REGULAR tier

## In Progress
- [x] `getAccountApprovalWhitelistDomains()` is not used, so need to delete the properties.
- [x] `licensedDomains and trialedDomains` properties can be removed and update places in slackService that use these.
- [x] Testing with slack
- [x] When user has a regular license and gets added to a company with trial license, then we don't downgrade the user


src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
```diff
@@ -244,6 +245,19 @@ public void sendApprovalEmail(UserDTO user) {
         sendEmailFromTemplate(user, MailType.APPROVAL);
     }
 
+    // This method should be used when we want to explain to the user that their
+    // company has a different license type and they've been approved to use that.
+    @Async
+    public void sendApprovalEmail(UserDTO userDTO, LicenseType registeredLicenseType, Company company) {
```
zhx828, line 251: Let's rename the method to reflect the aligning license message?

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -329,30 +329,6 @@ private LayoutBlock buildUserIdBlock(UserDTO userDTO) {
         return ContextBlock.builder().elements(elements).blockId(USER_ID.getId()).build();
     }
 
-    private List<LayoutBlock> buildWarningBlocks(UserDTO userDTO) {
```
zhx828, line 332: I think we still need the warning blocks here, just not he warning about the MSK user.

zhx828: The content needs to be updated to search the company info.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -537,9 +513,6 @@ private StaticSelectElement getLicenseTypeElement(UserDTO userDTO) {
         if (withDuplicateUserClarificationNote(userDTO, actionId)) {
             layoutBlocks.add(buildPlainTextBlock("We have sent a clarification email to the user asking why they registered multiple accounts.", DUPLICATE_USER_CLARIFICATION_NOTE));
         }
-        if (isMSKUser(userDTO)) {
```
zhx828, line 540: I think we still need this, but rather to apply to all regular commercial companies.

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -619,24 +626,36 @@ private void expireUserAccount(UserDTO userDTO){
...
+                return companies
+                    .stream()
+                    .filter(c -> {
+                        return c.getLicenseModel().equals(LicenseModel.REGULAR);
+                    })
+                    .findAny();
             }
         }
+        
         return Optional.ofNullable(null);
```
zhx828, line 659: why don't we return empty directly?

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -619,24 +626,36 @@ private void expireUserAccount(UserDTO userDTO){
...
+            if(companies.size() == 1){
+                Company company = companies.iterator().next();
+                // When there is only a micro tiered company, then we return empty optional.
+                if(company.getLicenseModel().equals(LicenseModel.REGULAR)){
+                    // If only a regular license model exists for domain, then auto approve user with company.
+                    return Optional.of(company);
+                }
+            }else if(companies.size() > 1){
+                // If there are mutliple companies with the domain, then we find a company with a regular license.
+                return companies
```
zhx828, line 650: This is a loophole. We should not have any domain be associated with multiple regular licensed companies. Though, this is really challenge to prevent in implementation. Let's discuss.

src/main/java/org/mskcc/cbio/oncokb/service/mapper/CompanyDomainMapper.java
```diff
@@ -9,13 +9,12 @@
...
-@Mapper(componentModel = "spring", uses = {CompanyMapper.class})
+@Mapper(componentModel = "spring", uses = {})
 public interface CompanyDomainMapper extends EntityMapper<CompanyDomainDTO, CompanyDomain> {
 
-    @Mapping(source = "company.id", target = "companyId")
-    CompanyDomainDTO toDto(CompanyDomain companyDomain);
 
-    @Mapping(source = "companyId", target = "company")
+    @Mapping(target = "companies", ignore = true)
+    @Mapping(target = "removeCompany", ignore = true)
```
zhx828, line 17: I'm curious what this is aboue.

src/main/resources/templates/mail/approvalAlignLicenseWithCompanyEmail.html
```diff
@@ -10,9 +10,10 @@
 
 <p th:text="'Thank you for your interest in the ' + ${license} + ' license for OncoKB.'"></p>
 
+<p th:text="'As a member of ' + ${companyName} + ', we are approving you for an OncoKB ' + ${companyLicenseType} + ' license which is available for no fee.'"></p>
```
zhx828, line 13:
```suggestion
<p th:text="'As a member of ' + ${companyName} + ', we are approving you for an OncoKB ' + ${companyLicenseType} + ' license.'"></p>
```

zhx828 (changes requested):
I will leave more product reviews before Monday.

calvinlu3:
Thanks for the feedbacks. Will get these addressed ASAP.

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
calvinlu3: I just made the adjustments to show the warning messages when domain is associated with multiple regular licensed companies. 
The domain buttons will be highlight with $warning and there will be an react-bootstrap Alert with a warning message.
Here is a screenshot:
image

zhx828 (changes requested):
- [x] For micro company, once the user is approved through slack, it should be associated with the company
- [x] `/companies/name/{name}` the company name could have special characters. Using the name in the path has limitation similar to https://github.com/oncokb/oncokb/issues/2764 Also maybe we should just create a generic company lookup method? `/company/lookup?name=***`
- [x] When a regular company is on trial, no need to send two messages to the channel. Only after user agrees with the trial, then send the activation message
- [x] Saw `org.hibernate.TransientObjectException:` when creating user with saved company
Screen Shot 2021-11-30 at 2 36 38 PM

zhx828 (approved):
Let's merge this pr and test the add-company branch directly.

PR #766, opened 2021-12-06
Disable User Registration

calvinlu3:
This is related to the [disable registration ticket](https://github.com/oncokb/oncokb/issues/2792)

#### Changes
- Created a `DisableEndpoint`  aop annotation that will send back `HttpStatus.Forbidden` to denote that the endpoint is disabled. 
   - Reference: https://www.baeldung.com/spring-aop-annotation
- Show an alert message on frontend
- Added property to set application to readonly mode.

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
@@ -1,6 +1,6 @@
 package org.mskcc.cbio.oncokb.web.rest;
 
-
+import org.mskcc.cbio.oncokb.aop.api.APIConditionallyEnabled;
 import org.mskcc.cbio.oncokb.config.application.ApplicationProperties;
 import org.mskcc.cbio.oncokb.domain.Token;
 import org.mskcc.cbio.oncokb.domain.User;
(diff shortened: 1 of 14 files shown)


src/main/java/org/mskcc/cbio/oncokb/aop/api/DisableEndpoint.java
```diff
@@ -0,0 +1,10 @@
+package org.mskcc.cbio.oncokb.aop.api;
+
+import java.lang.annotation.Target;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface DisableEndpoint {}
```
zhx828, line 10: What do you think if we call it `APIConditionallyEnabled`?

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -105,6 +106,7 @@ public AccountResource(UserRepository userRepository, UserService userService,
      * @throws EmailAlreadyUsedException {@code 400 (Bad Request)} if the email is already used.
      * @throws LoginAlreadyUsedException {@code 400 (Bad Request)} if the login is already used.
      */
+    @DisableEndpoint
```
zhx828, line 109: The reason I proposed the new name is at the place using this annotation, it's not so clear what it is for.

zhx828 (commented):
Just a minor comment.

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
calvinlu3: I agree that the name makes it clearer. Will make that change.

PR #774, opened 2021-12-30
Company updates

calvinlu3:
This pr fixes the following:
- [issue # 2848](https://github.com/oncokb/oncokb/issues/2848)
- [issue # 2847](https://github.com/oncokb/oncokb/issues/2847)
- [issue # 2845](https://github.com/oncokb/oncokb/issues/2845)

#### Changes
- `License Type`, `Business Contact`, and `Legal Contact` columns are removed from company users table
- On the `company page`, the company information can be displayed while the company user table and user dropdown are still waiting for data.
   - [x] Test this on beta.oncokb.org
- Allow admin to activate and deactivate user accounts in the company users table
  - Created a shared `UserStatusModal` component for Company Page and User Details Page
- [x] Update screenshots
   - Update `company` field to `companyName` in test data
   - Use the new User Details Page screenshot because we added a label `License Status` to the license status button selection.
- [x] Try fetching only user emails to reduce load time

PR #778, opened 2022-01-04
Fix issues from code review for company-update 

calvinlu3:
This PR addresses the comments in [company-updates pr](https://github.com/oncokb/oncokb-public/pull/774)

- Use remoteData to get status of api call
- Rename endpoint from /users/emails to /users/non-company-emails
- Remove call to UserRepository from CompanyService
- Fix the sorting issues brought up from weekly meeting.


src/main/webapp/app/pages/companyManagement/CompanyDetailsPage.tsx
```diff
@@ -34,7 +34,7 @@ export default class CompanyDetailsPage extends React.Component {
       id: 'companyUserCount',
       Header: <span className={styles.tableHeader}>Users</span>,
       minWidth: 60,
-      accessor: 'companyUserCount',
+      accessor: (data: CompanyDTO) => data.numberOfUsers,
```
zhx828, line 37: This is minor, you can change to `numberOfUsers` instead of calling the method.

src/main/webapp/app/pages/companyPage/CompanyPage.tsx
```diff
@@ -79,176 +81,174 @@ const LICENSE_STATUS_UPDATE_MESSAGES = {
...
+  }
+
+  @action.bound
+  updateCompany() {
+    if (!this.company.isComplete) {
+      notifyError(new Error('Cannot update when company is loading'));
+      return;
     }
+
+    /*eslint-disable no-console*/
```
calvinlu3, line 158: Removing this console log

src/main/webapp/app/pages/companyManagement/CompanyDetailsPage.tsx
calvinlu3: Thanks, also will change the accessor for companyName to `name` instead of calling the method.

src/main/webapp/app/pages/companyPage/CompanyPage.tsx
```diff
@@ -79,176 +81,170 @@ const LICENSE_STATUS_UPDATE_MESSAGES = {
...
-    } catch (error) {
-      notifyError(error);
+    },
+    default: [],
+  });
+
+  /* After the company is fetched, we will fetch the company users and the tokens */
+  @action.bound
+  getCompanyInfo(id: number) {
+    this.companyUsers = remoteData<UserDTO[]>({
```
zhx828, line 126: We should use `await` in the remoteData so when company is updated the companyUsers will be updated as well. This should be the case.

zhx828 (changes requested):
I think the main concern for these changes now is the usage of the remoteData. remoteData is a MobxPromise wrapper. We should use await/invoke of the remoteData to figure out the dependency of the api callings. Happy to discuss more during the standup.

zhx828:
@calvinlu3 could you also do me a favor to only show 5 users by default of the company table? Currently it's too long to move to the bottom of the page in a big company.

zhx828 (approved):

PR #779, opened 2022-01-04
Adjust page layout to have 13inch screen using the container-fluid

zhx828:
We want to maximize the page width on 13 inch screen, width 1440px


calvinlu3 (approved):
Tried it locally and seems fine to me.

PR #783, opened 2022-01-06
Show user ID & Update License Type visual based on company info

zhx828:
- show user id in the user page
- If the user has been associated with a company, we do not show license button but only show the current license
- this fixes https://github.com/oncokb/oncokb/issues/2864

### with company
Screen Shot 2022-01-06 at 11 35 09 AM

### without company
Screen Shot 2022-01-06 at 11 35 14 AM


calvinlu3 (approved):
This looks fine. Using a field to show the license type is better than the disabled buttons.

PR #789, opened 2022-01-10
Send approval email when regular user is automatically approved

calvinlu3:
This fixes issue: Users that were part of a regular, full licensed company were getting automatically approved and added to company, but did not receive an approval email.


zhx828 (approved):
LGTM

PR #791, opened 2022-01-12
Change logic for determining if user is on trial

calvinlu3:
This fixes [issue #2858](https://github.com/oncokb/oncokb/issues/2858)

- Users with at least one renewable token are considered regular, not trial.
- Set `notifyUserOnTrialCreation` to true when creating users in the CreateCompanyUsersPage. Currently, it's only sending the account is created email.


zhx828 (approved):
Looks good.

PR #793, opened 2022-01-13
Prevent user from logging in if trial license agreement is not accepted

calvinlu3:
This fixes [issue #2789](https://github.com/oncokb/oncokb/issues/2789)

If the user didn't accept the trial license agreement, then login fails. There will be a link to take the user to the trial agreement acceptance page at `/account/active-trial/finish?key=${key}`


src/main/java/org/mskcc/cbio/oncokb/web/rest/UserUUIDController.java
```diff
@@ -73,6 +82,26 @@ public UserUUIDController(TokenProvider tokenProvider, AuthenticationManagerBuil
             Token token = tokenProvider.createTokenForCurrentUserLogin(Optional.empty(), Optional.empty());
             uuid = token.getToken();
         }
+
+        Optional<UserDetailsDTO> userDetails = userDetailsService.findByUserIsCurrentUser();
+        if (userDetails.isPresent()) {
+            UserDetailsDTO ud = userDetails.get();
+            boolean userHasTrialKey =
```
zhx828, line 89: Try to use a method chaining. https://stackoverflow.com/questions/49380052/how-to-avoid-checking-for-null-values-in-method-chaining/49380103

src/main/webapp/app/components/login/LoginPage.tsx
```diff
@@ -66,6 +63,33 @@ export default class LoginPage extends React.Component<ILoginProps> {
...
+  get trialActivationLink() {
+    if (this.props.authenticationStore.loginError) {
+      const errorMessage = getErrorMessage(
+        this.props.authenticationStore.loginError
+      );
+      if (errorMessage.endsWith(UNAUTHORIZED_LICENSE_AGREEMENT_NOT_ACCEPTED)) {
+        const trialActivationKey = this.props.authenticationStore.loginError
+          .response?.body.trialActivationKey;
+        if (trialActivationKey) {
+          const to = `/account/active-trial/finish?key=${trialActivationKey}`;
```
zhx828, line 76: Could you use ACCOUNT_ACTIVE_TRIAL_FINISH? We probably should create a TrialActivationPageLink in UrlUtils.tsx.

zhx828 (approved):
I like the implementation. Just two minor comments.

PR #807, opened 2022-01-24
Pass hash parameters from ApiAccessPageNavTab to pages

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/2906

When redirecting to `/account/register`, the hash parameters were not passed along. 
Adding the `hash` field in the location object fixes this issue.


zhx828 (approved):
Thanks for the quick fix. 👍🏼

PR #814, opened 2022-01-27
Enable backend tests

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/2900

- Re-enabled backend test job on GitHub actions
   - ArchTest.java tests are disabled for now
- Removed `AuditResourceIT.java`, `UserJWTControllerIT.java`, and `TokenStatsResourceIT.java` since they're not in use
- Updated `ApiConditionallyEnabledAspect` to use `@Before` instead of `@Around` because advice doesn't need to run after the method is ran. It was causing some issues with the response status (returning HTTP status 200).


src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -247,6 +248,9 @@ public User registerUser(UserDTO userDTO, String password) {
         userRepository.findOneByEmailIgnoreCase(userDTO.getEmail()).ifPresent(existingUser -> {
             throw new EmailAlreadyUsedException();
         });
+        userRepository.findOneByLogin(userDTO.getLogin().toLowerCase()).ifPresent(existingUser -> {
+            throw new LoginAlreadyUsedException();
+        });
```
calvinlu3, line 253: A user's login is the same as their email, so the first check should be sufficient. However, the login check is added in case that is changed.

zhx828 (approved):
Looking good 👍🏼

PR #827, opened 2022-02-22
Allow fetching 1M records in token stats pagination for backing up to s3

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/2954


benzuzu (approved):
Looks good!

PR #836, opened 2022-03-11
Disable recaptcha

zhx828:
The tool becomes so aggressive, even after validating, it keeps showing up on every page redirect.


calvinlu3 (approved):
Noted.

PR #840, opened 2022-03-18
Allow users to generate a new token.

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3035

### Changes
- User tokens are sorted by their lengths, so the last token (longest) cannot be deleted.
- Users can generate new token if they have only 1 token. 
- When user generates a new token, the old token will expire in `min(7 days, current expiration)`.

### New Changes
- User can delete any token now. The token's expiration will be applied to another token.

Other:
- Fixed tooltip not showing on disabled buttons.
- `AccountMessage` shows the trial about to expire banner at top of page even when the user logged out, so that is fixed.


zhx828 (approved):
Two minor comments.

zhx828 (approved):

PR #846, opened 2022-03-31
Update token stats before deleting token

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3058

Before deleting the token, we associate the token's token stats with another token.


src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -307,13 +312,24 @@ public void deleteToken(@RequestBody Token token) throws AuthenticationException
...
+                    tokens = tokens.stream().filter(t -> !t.getId().equals(deleteTokenId)).collect(Collectors.toCollection(ArrayList::new));
                     Token longestToken = tokens.stream().max(Comparator.comparing(Token::getExpiration)).get();
                     if (timestamp.isAfter(longestToken.getExpiration())) {
                         longestToken.setExpiration(timestamp);
                         tokenService.save(longestToken);
                     }
+
+                    // Find the token stats for the token to be deleted and associate those stats with the longest token.
+                    tokenStatsService
+                        .getAllTokenStatsByTokenId(deleteTokenId)
```
zhx828, line 325: Could you update the token id on sql level instead of getting all tokens stats into cache?

src/main/java/org/mskcc/cbio/oncokb/service/TokenService.java
```diff
@@ -55,10 +55,11 @@
...
     /**
-     * Delete the "id" token.
-     *
-     * @param id the id of the entity.
+     * Deletes the oldToken and associates its token stats with a new token.
+     * @param oldToken token to be deleted
+     * @param newToken token to associate the old token's token stats
      */
-    void delete(Long id);
+    void delete(Token oldToken, Token newToken);
```
zhx828, line 64: I don't think we should change the signature. We can have the delete method to randomly pick a token to update to.

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -307,13 +309,14 @@ public void deleteToken(@RequestBody Token token) throws AuthenticationException
...
                     Instant timestamp = token.getExpiration();
-                    tokenService.delete(token.getId());
-                    tokens = tokenService.findByUser(token.getUser());
+                    Long deleteTokenId = token.getId();
+                    tokens = tokens.stream().filter(t -> !t.getId().equals(deleteTokenId)).collect(Collectors.toCollection(ArrayList::new));
                     Token longestToken = tokens.stream().max(Comparator.comparing(Token::getExpiration)).get();
                     if (timestamp.isAfter(longestToken.getExpiration())) {
                         longestToken.setExpiration(timestamp);
                         tokenService.save(longestToken);
                     }
```
zhx828, line 318: Should we move these all logics to TokenService delete method?

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/service/impl/TokenServiceImpl.java
```diff
@@ -106,10 +118,44 @@ public void increaseTokenUsage(Long id, int increment) {
         return tokenRepository.findAllExpiresBeforeDate(date);
     }
 
+    @Override
+    public void expireToken(Token token) {
+        token.setExpiration(Instant.now());
+        save(token);
```
calvinlu3, line 124: Moved the `expireToken` method from `TokenProvider` to `TokenService` to avoid the circular dependencies.

zhx828 (approved):

PR #847, opened 2022-04-07
Allow multiple message banners on top && Show AACR Poster Session banner

zhx828:


src/main/webapp/app/components/userMessager/UserMessage.tsx
```diff
@@ -116,62 +135,46 @@ export default class UserMessage extends React.Component<UserMessageProps> {
...
-      !!this.shownMessage
-    );
+    return this.props.show && this.messages.length > 0;
   }
 
   render() {
     if (this.showBeVisible) {
-      return (
+      return toJS(this.messages).map(message => (
         <div className={styles.message}>
```
calvinlu3, line 165: Do we need a `key` here?

calvinlu3 (approved):

src/main/webapp/app/components/userMessager/UserMessage.tsx
zhx828: Good point, will update

PR #850, opened 2022-04-20
Use class loader to fetch file under the resource folder

zhx828:
This fixes issue that email template cannot be fetched in production when path changes.
This fixes https://github.com/oncokb/oncokb/issues/3079

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
@@ -36,7 +36,9 @@ import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Service;
 import org.thymeleaf.context.Context;
 
+import java.io.File;
 import java.io.IOException;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
@@ -102,7 +104,7 @@ public class SlackService {
 
     @Async
     public void sendApprovedConfirmation(UserDTO userDTO, Company company) {
-        String text = 
+        String text =
             String.format(
                 "%s has been approved to use %s's %s license and notified automatically.",
                 userDTO.getEmail(),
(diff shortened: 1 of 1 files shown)


src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -815,10 +816,10 @@ private LayoutBlock buildPlainTextBlock(String text, BlockId blockId) {
         return SectionBlock.builder().text(PlainTextObject.builder().text(text).build()).blockId(blockId.getId()).build();
     }
 
-    private String getStringFromTextFile(String filePath) {
+    private String getStringFromResourceTemplateMailTextFile(String fileName) {
         StringBuilder sb = new StringBuilder();
 
-        try (Stream<String> stream = Files.lines(Paths.get(filePath), StandardCharsets.UTF_8)) {
+        try (Stream<String> stream = Files.lines(Paths.get(getClass().getClassLoader().getResource("templates/mail/" + fileName).getPath()), StandardCharsets.UTF_8)) {
```
calvinlu3, line 822: `getResources(...)` could return `null` if the the resource is not found. Is it necessary to add a null check?

calvinlu3 (approved):

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
benzuzu: It may be good to add in case template names are changed. 

Also, is `/src/main/resources/` not needed in the file path?

benzuzu (approved):

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
zhx828: @benzuzu yup, the method `getClass().getClassLoader().getResource` will locate the file in the resource folder. In the build jar/war file. src/main/resources is no longer available.

PR #854, opened 2022-05-03
Minor updates

zhx828:
This is related to https://github.com/oncokb/oncokb/issues/3087

### DIFF
--- a/src/main/webapp/app/shared/utils/Utils.tsx
+++ b/src/main/webapp/app/shared/utils/Utils.tsx
@@ -687,7 +687,13 @@ export function getAlterationName(
     typeof alteration === 'string' ? alteration : alteration.name;
   const isCategoricalAlt = isCategoricalAlteration(alt);
   const hasExclusionArm = alt.includes('{');
-  if (!showNameDiff || alt === name || isCategoricalAlt || hasExclusionArm) {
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/shared/utils/Utils.tsx
```diff
@@ -687,7 +687,13 @@ export function getAlterationName(
     typeof alteration === 'string' ? alteration : alteration.name;
   const isCategoricalAlt = isCategoricalAlteration(alt);
   const hasExclusionArm = alt.includes('{');
-  if (!showNameDiff || alt === name || isCategoricalAlt || hasExclusionArm) {
+  if (
+    !showNameDiff ||
+    alt === name ||
+    isCategoricalAlt ||
+    hasExclusionArm ||
+    name.includes(alt)
```
calvinlu3, line 695: What does this condition do?

calvinlu3 (approved):

src/main/webapp/app/shared/utils/Utils.tsx
zhx828: To deal with when alteration is `Promoter` and name is `Promoter Mutation`, then we should `Promoter Mutation` only.

PR #857, opened 2022-05-05
Prod yaml prop file should only include ones with value

zhx828:
For other props that without value, they are considered as optional variable, and then should not be included.

This is related to https://github.com/oncokb/oncokb/issues/3099


calvinlu3 (approved):
Noted this change.

PR #858, opened 2022-05-05
Updates to smartsheet integration

zhx828:
- Do not send MSK users to the sheet
- For users created by the admin, these accounts need to be sent to ROC sheet.
- Do not send user that we sent academic email clarification to the sheet
- Give option to opt-out ROC review on user creation page

This fixes to https://github.com/oncokb/oncokb/issues/3088


calvinlu3 (approved):
LGTM. There are two failing tests.

PR #871, opened 2022-06-14
Update Usage Analysis Page

benzuzu:
- Usage Analysis page now calls data from the correct s3 path

- Removed old usage analysis cronjob. Redundant version of the pipeline project.

Checklist:

First step:
- [x] Update file path to populate the front end table
- [x] Add new daily usage from monthly analysis files
- [x] Calculate resource detail on the front end and populate table

Second step:
- [x] Add calendar to daily user usage to query specific dates
- [x] Add a new tab "total resources" to see total usage over all endpoints

Third step:
- [x] Add usage table to user page
- [x] Add usage table to company page for only company users
- [ ] Analyze company usage as well as individual usage (can be done in separate pr)

this fixes https://github.com/oncokb/oncokb/issues/3152 and oncokb/oncokb#3189 and oncokb/oncokb#3213


zhx828 (changes requested):
Few comments added, could you also update the pr with the TODO items we discussed during the meeting? Thanks!

src/main/webapp/app/components/oncokbTable/OncoKBTable.tsx
```diff
@@ -47,20 +48,37 @@ export default class OncoKBTable<T> extends React.Component<
     });
   }
 
+  SearchBar = () => {
```
zhx828, line 51: We would not recommend having a component within a class component. You could consider moving the logic to a functional component.

src/main/webapp/app/components/oncokbTable/OncoKBTable.tsx
```diff
@@ -47,20 +48,37 @@ export default class OncoKBTable<T> extends React.Component<
...
-              <input
-                onChange={(event: any) => {
-                  this.searchKeyword = event.target.value.toLowerCase();
-                }}
-                className="form-control input-sm"
-                type="text"
-                placeholder="Search ..."
-              />
+        {this.props.filters === undefined ? (
+          <this.SearchBar />
```
zhx828, line 74: I think we need to consider here is that the div with row className should always be there. If the filters is not null, then render the filters. If the search is enabled, then render the search bar.

zhx828 (changes requested):

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+  currentDate: string;
+  currentFromDate: string;
+  currentToDate: string;
+  fromDate: (newDate: string) => void;
+  toDate: (newDate: string) => void;
+  filterToggled: (filterActive: boolean) => void;
+};
+
+export const UsageAnalysisCalendarButton: React.FunctionComponent<CalendarButtonProps> = props => {
+  function handleFromDayChange(day: any) {
```
zhx828, line 35: We should really try to avoid using `any`. It defeats the purpose of using TS. Could you take a look and update if possible? This includes the type for `filters` below.

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  function handleToDayChange(day: any) {
+    props.toDate(day);
+    props.filterToggled(true);
+  }
+
+  function handleResetDays() {
+    props.fromDate('Select Date');
+    props.toDate('Select Date');
```
zhx828, line 50: I guess this could be a constant?

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <Container>
+          <div>
+            <Row>
+              <Col>From:</Col>
+            </Row>
+          </div>
+          <DayPickerInput
+            value={
+              props.currentFromDate === undefined ||
+              props.currentFromDate === 'Select Date'
```
zhx828, line 71: If the type defines as string and is required field, don't check whether it's undefined. If the expectation of `currentFromDate` is string || undefined, please update the type.

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+        as={InputGroup.Append}
+        title={<i className={classnames('fa fa-calendar')}></i>}
+        id="time-select-dropdown"
+        className="ml-3"
+      >
+        <Container>
+          <div>
+            <Row>
+              <Col>From:</Col>
+            </Row>
```
zhx828, line 66: I guess my question here will be, do you really need to use multiple rows and columns? I don't see you need to do any alignment of columns?

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+                border: 'none',
+              },
+            }}
+            dayPickerProps={{
+              initialMonth:
+                props.currentToDate === undefined
+                  ? moment().toDate()
+                  : getMomentInstance(props.currentToDate).toDate(),
+            }}
+            style={{ margin: '0 auto' }}
```
zhx828, line 115: What the purpose of specifying the margin?

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,132 @@
...
+            formatDate={MomentLocaleUtils.formatDate}
+            parseDate={MomentLocaleUtils.parseDate}
+          />
+          <Row className={'mt-1'}>
+            <Button
+              size={'sm'}
+              style={{ margin: '0 auto', zIndex: 0 }}
```
zhx828, line 122: What happens if you don't specify zIndex?

src/main/webapp/app/components/oncokbTable/OncoKBTable.tsx
```diff
@@ -50,17 +51,34 @@ export default class OncoKBTable<T> extends React.Component<
...
-            <div className="ml-auto">
-              <input
-                onChange={(event: any) => {
-                  this.searchKeyword = event.target.value.toLowerCase();
-                }}
-                className="form-control input-sm"
-                type="text"
-                placeholder="Search ..."
-              />
+        {this.props.filters === undefined && this.props.disableSearch ? (
```
zhx828, line 54: What's the purpose of adding this additional check? Prevent rendering extra div?

src/main/webapp/app/pages/usageAnalysisPage/UserUsageDetailsTable.tsx
```diff
@@ -47,43 +56,103 @@ export default class UserUsageDetailsTable extends React.Component<
...
-        ? USGAE_ALL_TIME_KEY
-        : USAGE_DETAIL_TIME_KEY
+        ? USAGE_ALL_TIME_KEY
+        : this.timeTypeToggleValue === ToggleValue.RESULTS_BY_MONTH
+        ? USAGE_DETAIL_TIME_KEY
+        : USAGE_DAY_DETAIL_TIME_KEY
     );
+    if (this.filterToggled) {
+      if (this.timeTypeToggleValue === ToggleValue.RESULTS_BY_MONTH) {
+        data = data?.filter(resource => {
```
zhx828, line 68: Again, data type check?

src/main/webapp/app/pages/usageAnalysisPage/UserUsageDetailsTable.tsx
```diff
@@ -47,43 +56,100 @@ export default class UserUsageDetailsTable extends React.Component<
...
+        const cumulativeData: UsageRecord[] = [];
+        data.forEach(value => {
+          if (!cumulativeData.find(value1 => value1.time === value.time)) {
+            cumulativeData.push({
+              resource: 'ALL',
+              usage: 0,
+              time: value.time,
+            });
+          }
+          // @ts-ignore
```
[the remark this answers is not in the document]
zhx828: We should not use ts-ignore unless it's absolute necessary.

zhx828 (changes requested):
Mostly my complains are about the typescript usage which is super important for us. Let's spend some time fixing these.

src/main/webapp/app/components/oncokbTable/OncoKBTable.tsx
benzuzu: yes, the mt-2 row div creates 8-pixel wide blank row due to the margin. @zhx828

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
benzuzu: zIndex pushes the reset button all the way to the back so it doesn't appear above the calendar like so:

Screen Shot 2022-07-23 at 10 03 47 AM

src/main/webapp/app/pages/usageAnalysisPage/ResourceUsageDetailsTable.tsx
```diff
@@ -37,23 +37,28 @@ export default class ResourceUsageDetailsTable extends React.Component<
...
+          defaultValue={this.timeTypeToggleValue}
+          toggleValues={[
+            ToggleValue.RESULTS_IN_TOTAL,
+            ToggleValue.RESULTS_BY_MONTH,
+          ]}
+          handleToggle={this.handleTimeTypeToggleChange}
+        />
+      </Row>
+    );
+  };
```
zhx828, line 53: Can we move this to a standalone functional component?

zhx828 (changes requested):
Two minor comments, once fix we can merge.

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+  currentToDate: string | undefined;
+  fromDate: (newDate: string) => void;
+  toDate: (newDate: string) => void;
+  filterToggled: (filterActive: boolean) => void;
+};
+
+const DATE_PLACEHOLDER = 'Select Date';
+
+export const UsageAnalysisCalendarButton: React.FunctionComponent<CalendarButtonProps> = props => {
+  function handleFromDayChange(day: Date) {
```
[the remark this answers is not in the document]
benzuzu: You can only clear the date selection with the reset button, which calls the `handleResetDays()` function. So the date will always be defined.

src/main/webapp/app/pages/usageAnalysisPage/ResourceUsageDetailsTable.tsx
benzuzu: @zhx828 I don't think we can move it outside of the class component as it needs to be able to modify the `timeTypeToggleValue` which is a property of the class component and changes the table. Is this what you meant for me to do?

zhx828 (approved):

zhx828:
@benzuzu could you check what causes the screenshot changes?

src/main/webapp/app/components/oncokbTable/OncoKBTable.tsx
```diff
@@ -50,17 +51,34 @@ export default class OncoKBTable<T> extends React.Component<
...
+          <></>
+        ) : (
+          <div className="mt-2 row">
```
benzuzu, line 57: @zhx828 the cause of the screenshot test failure is here I believe. Since a margined row is created regardless of whether there are filters, there will be a few extra pixels above the search bar on other tables without filters. The old logic was that the margined row was only created if the filters were present, but I believe we decided to change it to creating a row regardless and adding filters and the search bar after the row is created.

zhx828 (changes requested):

src/main/webapp/app/pages/companyPage/CompanyPage.tsx
```diff
@@ -595,6 +627,108 @@ export default class CompanyPage extends React.Component<
...
+                          <OncoKBTable
+                            data={this.users.result}
+                            columns={[
+                              {
+                                id: 'userEmail',
+                                Header: <span>Email</span>,
+                                accessor: 'userEmail',
+                                minWidth: 200,
+                                onFilter: (data: UserOverviewUsage, keyword) =>
+                                  filterByKeyword(data.userEmail, keyword),
```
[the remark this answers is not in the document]
benzuzu: I think it still needs to be defined otherwise it will not show up on the table.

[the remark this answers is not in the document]
benzuzu: This configures the search bar. I think you have to manually filter by keyword.

src/main/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisController.java
```diff
@@ -57,74 +69,127 @@ private JSONObject requestData(String file)
...
+        do {
+            String month = TimeUtil.getCurrentNYTime().minus(monthsBack, ChronoUnit.MONTHS).format(DateTimeFormatter.ofPattern("yyyy-MM"));
+            monthSummary = requestData(MONTH_USERS_USAGE_SUMMARY_FILE_PREFIX + month + FileExtension.JSON_FILE.getExtension());
+            if (monthSummary != null) {
+                monthSummaries.add(monthSummary);
+            }
+            monthsBack++;
+        } while (monthsBack < 12);
+
+        Long id = Long.parseLong(userId);
```
zhx828, line 89: Let's do a type check, otherwise here throws exception when userId is null.

src/main/webapp/app/components/oncokbTable/HeaderConstants.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const emailHeader = <span>Email</span>;
+export const usageHeader = <span>Usage</span>;
+export const resourceHeader = <span>Resource</span>;
+export const endpointHeader = <span>Most frequently used endpoint</span>;
+export const noPrivateEndpointHeader = (
+  <span>Most frequently used endpoint(only public)</span>
+);
+export const timeHeader = <span>Time</span>;
+export const operationHeader = <span>Details</span>;
```
zhx828, line 12: - We should at minimum Uppercase the first letter since they are components.
- I would not prefer wrap constant string in a span component. It's just unnecessary <span> when you wrap it in other tags
- The table header could be string

src/main/webapp/app/pages/userPage/UserPage.tsx
```diff
@@ -126,6 +142,56 @@ export default class UserPage extends React.Component<IUserPage> {
...
+      Object.keys(yearSummary).forEach(resourceEntry => {
+        yearUsage.push({
+          resource: resourceEntry,
+          usage: yearSummary[resourceEntry],
+          time: USAGE_ALL_TIME_VALUE,
+        });
+      });
+      result.set(USAGE_ALL_TIME_KEY, yearUsage);
+
+      const monthSummary = this.userUsage.summary.month;
```
zhx828, line 163: When a user has no usage, nor id is invalid, you got not summary data. Here will throw exception causing the promise never been resolved. Thereafter the loading state never changes.

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+  currentFromDate: string | undefined;
+  currentToDate: string | undefined;
+  fromDate: (newDate: string) => void;
+  toDate: (newDate: string) => void;
+  filterToggled: (filterActive: boolean) => void;
+};
+
+const DATE_PLACEHOLDER = 'Select Date';
+
+export const UsageAnalysisCalendarButton: React.FunctionComponent<CalendarButtonProps> = props => {
```
zhx828, line 36: - [ ] Can we do not close the window after selecting the From?
- [ ] We should indicate somehow when the calendar filter is selected. Maybe change the background to dark blue like when you hover over?

zhx828 (changes requested):

src/main/webapp/app/pages/usageAnalysisPage/UserUsageDetailsPage.tsx
```diff
@@ -43,24 +44,42 @@ export default class UserUsageDetailsPage extends React.Component<{
       const result = new Map<string, UsageRecord[]>();
```
zhx828, line 44: Could you do me a favor to fix the line `<Link to={`${PAGE_ROUTE.USER}/${this.user.userEmail}`}>`? I can't leave a comment there as there is no change comparing to master. But the link is wrong, it populates as /users/:login/useremail, but it should be /users/useremail

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,134 @@
...
+import DayPickerInput from 'react-day-picker/DayPickerInput';
+import 'react-day-picker/lib/style.css';
+import moment from 'moment';
+import { APP_LOCAL_DATE_FORMAT } from 'app/config/constants';
+import MomentLocaleUtils from 'react-day-picker/moment';
+import { DefaultTooltip } from 'cbioportal-frontend-commons';
+import pluralize from 'pluralize';
+
+type CalendarButtonProps = {
+  currentDate: string | undefined;
```
zhx828, line 14: not used

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,134 @@
...
+  currentFromDate: string | undefined;
+  currentToDate: string | undefined;
+  currentToggleState: boolean;
```
zhx828, line 17: Shouldn't `currentToggleState` be calculated by currentFromDate and currentToDate?

src/main/webapp/app/components/calendarButton/UsageAnalysisCalendarButton.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { InputGroup, Button, Dropdown, Container } from 'react-bootstrap';
+import classnames from 'classnames';
+import { getMomentInstance } from 'app/shared/utils/Utils';
+import DayPickerInput from 'react-day-picker/DayPickerInput';
+import 'react-day-picker/lib/style.css';
+import moment from 'moment';
+import { APP_LOCAL_DATE_FORMAT } from 'app/config/constants';
+import MomentLocaleUtils from 'react-day-picker/moment';
+import { DefaultTooltip } from 'cbioportal-frontend-commons';
+import pluralize from 'pluralize';
```
zhx828, line 11: Not used

zhx828 (changes requested):

zhx828 (approved):

PR #872, opened 2022-06-28
Remove inactivated trial account's information when user license changes to regular

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3174

Reason: A user has not activated their trial license will still have their trial account activation `key`, so this PR (https://github.com/oncokb/oncokb-public/pull/793) prevents the user from logging in. However, if the user later gets added to a regular company, that information should be cleared or else the user not be able to login.

- Fixed trial account agreement link not showing when user who hasn't agreed to terms tries to login. This was due to the OncoKB text replaced with the trademark version in the frontend, but was not reflected in the backend error constant.
- Remove trial account information when a user's license changes to regular or is expired if the trial was never activated. 
- Added a few extra test cases

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
@@ -299,11 +299,6 @@ public class UserService {
         return newUser;
     }
 
-    public boolean trialAccountActivated(UserDTO userDTO) {
-        List<Token> tokens = tokenService.findByUser(userMapper.userDTOToUser(userDTO));
-        return !tokens.stream().filter(token -> token.isRenewable()).findAny().isPresent();
-    }
-
     public boolean trialAccountInitiated(UserDTO userDTO) {
         if (
             userDTO.getAdditionalInfo() == null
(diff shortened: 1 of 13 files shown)


src/test/java/org/mskcc/cbio/oncokb/service/CompanyServiceIT.java
```diff
@@ -191,23 +193,57 @@ public void assertThatUserIsRegularWhenCompanyOnRegularLicense() {
...
+        // Change license to regular
+        CompanyVM updatedCompanyVM = createCompanyVM(existingCompany);
+        updatedCompanyVM.setLicenseStatus(LicenseStatus.REGULAR);
+        companyService.updateCompany(updatedCompanyVM);
+
+        // Check assumptions
+        Optional<User> optionalUserPost = userRepository.findOneById(newUser.getId());
+        UserDTO userDTOPost = userMapper.userToUserDTO(optionalUserPost.get());
+        assertThat(userDTOPost.getAdditionalInfo()).isNotNull();
+        assertThat(userDTOPost.getAdditionalInfo().getTrialAccount()).isNotNull();
```
zhx828, line 228: Could you also do a check on the activation date?

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -701,13 +743,33 @@ private boolean userHasValidTrialActivation(UserDTO userDTO){
             .isPresent();
     }
 
-    private boolean userOnTrial(UserDTO userDTO) {
+    /**
+     * Checks whether the user is currently on trial. A user is considered
+     * to be on trial if they have at least one renewable token.
```
zhx828, line 748:
```suggestion
     * to be on trial if they have no renewable token.
```

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -646,17 +659,41 @@ public Optional<UserDTO> updateUserWithCompanyLicense(UserDTO userDTO, Company c
...
+            userDTO.getAdditionalInfo().setTrialAccount(null);
+            Optional<UserDetails> userDetails = userDetailsRepository.findOneByUser(userMapper.userDTOToUser(userDTO));
+            if (userDetails.isPresent()) {
+                UserDetails ud = userDetails.get();
+                String newAdditionalInfoString = null;
+                // Preserve userCompany information if present
+                if (userDTO.getAdditionalInfo().getUserCompany() != null) {
+                    newAdditionalInfoString = new Gson().toJson(userDTO.getAdditionalInfo());
+                }
+                ud.setAdditionalInfo(newAdditionalInfoString);
```
zhx828, line 692: UserCompany is not guaranteed to be the only data element besides the trial account in the additional info in the future. I would just leave the additional info there.

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
calvinlu3: I removed the code for checking `userCompany` and just set the `trialAccount` to `null`. Then I added a util function to check if the `additionalInfo` object is empty.

src/main/java/org/mskcc/cbio/oncokb/util/ObjectUtil.java
```diff
@@ -0,0 +1,19 @@
...
+     * Checks if a AdditionalInfoDTO is empty or null. AdditionalInfoDTO is empty if its fields are all empty as well.
+     * 
+     * @param additionalInfoDTO the user's additionalInfoDTO
+     * @return true if null or empty, otherwise false
+     */
+    public static boolean isUserAdditionalInfoEmpty(AdditionalInfoDTO additionalInfoDTO) {
+        if (additionalInfoDTO == null) {
+            return true;
+        }
+        return additionalInfoDTO.getTrialAccount() == null && additionalInfoDTO.getUserCompany() == null;
```
zhx828, line 17: This still does not give any warning nor error if properties of AdditionalInfo changes.
Can we do Gson toJson and check the String is "{}"? Or similar idea, sounds a bit hacky though.

zhx828: Then we can even take any Object and convert it to Json? name could be `isObjectEmpty`?

zhx828: Gson might throw exception though, might need to handle that. I'm not sure.

PR #878, opened 2022-07-27
Use fdaLevel from evidence table to generate FdaAlteration info

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3198


calvinlu3 (approved):
The screenshots need to be updated, otherwise looks good.

PR #880, opened 2022-08-10
Allow admin to fully delete the user

zhx828:
We will delete all information related to the user, includes the following
- user mails
- tokens
- token stats
- user details

This fixes https://github.com/oncokb/oncokb/issues/2098

The following section will be added to the user page.
Screen Shot 2022-08-10 at 4 54 15 PM


calvinlu3 (approved):
LGTM

PR #885, opened 2022-08-31
Expires token when user is inactivated

scarrero4660:
Expires token immediately when user is inactivated from user details page. This fixes https://github.com/oncokb/oncokb/issues/3221

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
@@ -420,7 +420,11 @@ public class UserService {
 
         if (updatedUserDTO.isPresent() && updatedUserDTO.get().isActivated()) {
             generateTokenForUserIfNotExist(updatedUserDTO.get(), Optional.empty(), Optional.empty());
+
+        } else if (updatedUserDTO.isPresent() && !updatedUserDTO.get().isActivated()) {
+            expireUserAccount(userDTO);
         }
+
         return updatedUserDTO;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -701,17 +704,13 @@ public Optional<UserDTO> updateUserWithCompanyLicense(UserDTO userDTO, Company c
...
-            tokens.forEach(token -> {
-                tokenService.expireToken(token);
-            });
+        if (userHasUnactivatedTrial(userDTO)) {
+            clearTrialAccountInformation(userDTO);
         }
+        List<Token> tokens = tokenService.findByUser(userMapper.userDTOToUser(userDTO));
+        tokens.forEach(token -> {
+            tokenService.expireToken(token);
+        });
```
zhx828, line 713: Why we want to make this change?

scarrero4660: Calling updateUser would result in an infinite loop because expireToken is called in updateUser. It is unnecessary because the user is already checked and updated before it is passed in to expireToken in the updateUser method

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -420,7 +420,10 @@ public Optional<UserDTO> updateUser(UserDTO userDTO) {
 
         if (updatedUserDTO.isPresent() && updatedUserDTO.get().isActivated()) {
             generateTokenForUserIfNotExist(updatedUserDTO.get(), Optional.empty(), Optional.empty());
+        } else if (updatedUserDTO.isPresent() && !updatedUserDTO.get().isActivated()) {
+            expireUserAccount(userDTO);
         }
```
zhx828, line 425: These two can be merged together.

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
zhx828: But this method needs to inactivate the userDTO if it hasn't. Let's add a conditional check on activated status, it's still true, we need to set it to false and call updateUser.

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
scarrero4660: This method does still inactivate the userDTO before the lines I added. I will add the conditional check

zhx828 (approved):
LGTM, thanks!

PR #896, opened 2022-09-12
Improve User Details Page and Company Page's User Dropdown Performances

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/2342

- Cache `getAllManagedUsers` method in `UserService.java`
- Update `usersToUserDTOs` method to fetch user + user details in bulk
- Update `getNonCompanyUserEmails` method in `UserService.java` to use custom query
   - The original query takes about 7-8 seconds, but the custom query takes ~50ms.

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
@@ -22,7 +22,6 @@ import org.mskcc.cbio.oncokb.service.dto.CompanyDTO;
 import org.mskcc.cbio.oncokb.service.dto.UserDTO;
 import org.mskcc.cbio.oncokb.service.mapper.UserMapper;
 import org.mskcc.cbio.oncokb.service.mapper.CompanyMapper;
-import org.mskcc.cbio.oncokb.service.mapper.UserDetailsMapper;
 import org.mskcc.cbio.oncokb.util.ObjectUtil;
 import org.mskcc.cbio.oncokb.util.StringUtil;
 import org.mskcc.cbio.oncokb.web.rest.errors.LoginAlreadyUsedException;
@@ -35,10 +34,12 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.actuate.audit.AuditEvent;
 import org.springframework.cache.CacheManager;
 import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageImpl;
 import org.springframework.data.domain.Pageable;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
+import org.springframework.cache.annotation.Cacheable;
 
 import javax.validation.constraints.NotNull;
 import java.time.Instant;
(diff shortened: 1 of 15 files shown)


calvinlu3:
@zhx828 , I had to put `findAllUsersWithUserDetailsByUsersIn(...)` in `UserDetailsService` instead of `UserService` due to circular dependencies.

src/main/java/org/mskcc/cbio/oncokb/service/impl/UserDetailsServiceImpl.java
```diff
@@ -84,4 +85,17 @@ public Optional<UserDetailsDTO> findByUserIsCurrentUser() {
...
+
+    @Transactional(readOnly = true)
+    public List<UserAndUserDetailsPair> findAllUsersWithUserDetailsByUsersIn(List<User> users) {
+        List<Object[]> usersWithDetails = userDetailsRepository.findAllUsersWithUserDetailsByUsersIn(users);
+        return usersWithDetails
+            .stream()
+            .map(u -> {
+                User user = (User) u[0];
+                UserDetails userDetails = (UserDetails) u[1];
+                return new UserAndUserDetailsPair(user, userDetails);
```
zhx828, line 97: I just realize that isn't this UserDTO? Let's return the UserDTO directly?

zhx828 (changes requested):
We need to address the circular dependency issue in the future. The mapper files probably should leave the repository/service out.

src/main/java/org/mskcc/cbio/oncokb/service/impl/UserDetailsServiceImpl.java
calvinlu3: Yeah it is very similar. The reason is that [we add the companyDTO](https://github.com/oncokb/oncokb-public/blob/master/src/main/java/org/mskcc/cbio/oncokb/service/mapper/UserMapper.java#L40) when mapping the user to userDTO. If I used the UserDTO, then I cannot map to a CompanyDTO.

zhx828: Are you adding it here? https://github.com/oncokb/oncokb-public/pull/896/files#diff-ccdbff722436ab7c2f7d0a2d7ecc10a40b0b25de777c602ff46dce2e588d8009R41

calvinlu3: Yeah, I am adding the companyDTO at that line. If I returned a UserDTO, then I'm not able to add that line in the UserMapper.

zhx828: Should we add a new method in UserDTO, a copy of https://github.com/oncokb/oncokb-public/blob/master/src/main/java/org/mskcc/cbio/oncokb/service/mapper/UserMapper.java#L36 but accepting UserDetails as second parameter?

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -508,8 +507,24 @@ public void changePassword(String currentClearTextPassword, String newPassword)
...
+                User user = (User) u[0];
+                UserDetails userDetails = (UserDetails) u[1];
+                return userMapper.userToUserDTO(user, userDetails);
+            })
+            .collect(Collectors.toList());
+    }
+
+    @Transactional(readOnly = true)
+    @Cacheable(cacheResolver = "userCacheResolver", key = "#root.methodName")
     public Page<UserDTO> getAllManagedUsers(Pageable pageable) {
```
zhx828, line 524: Could you do me a favor checking whether we can get rid of the threshold https://github.com/oncokb/oncokb-public/blob/master/src/main/webapp/app/pages/userManagement/UserDetailsPage.tsx#L78? So the pageable will be null here I assume? If not, let's change that to a bigger size? 10K? @calvinlu3

calvinlu3: The Pageable object populates with default parameters when we get rid of the threshold. The default size is 20.
I'm increasing the page size to 10k.

zhx828 (approved):

PR #898, opened 2022-09-13
Add new team member

zhx828:


src/main/webapp/app/pages/teamPage/TeamPage.tsx
```diff
@@ -61,6 +61,13 @@ export const TeamPage = () => {
       faculty: false,
       showCOI: true,
     },
+    {
+      firstName: 'Stephanie',
```
zhx828, line 65: @scarrero4660 sorry, I forgot to update the website at the time of joining. Could you double check your info?

scarrero4660: No problem. My only note is that I have a BA not a BSc

zhx828: Sounds good, let me update.

PR #899, opened 2022-09-30
Recaptcha enterprise with endpoint protection and server side validation

scarrero4660:
This fixes https://github.com/oncokb/oncokb/issues/3259,  https://github.com/oncokb/oncokb/issues/3261,  https://github.com/oncokb/oncokb/issues/3034 and https://github.com/oncokb/oncokb/issues/3358


src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -360,6 +360,9 @@ public User createUser(UserDTO userDTO, Optional<Integer> tokenValidDays, Option
         userDetails.setCompanyName(userDTO.getCompanyName());
         userDetails.setCity(userDTO.getCity());
         userDetails.setCountry(userDTO.getCountry());
+        if(userDTO.getAdditionalInfo() != null) {
+            userDetails.setAdditionalInfo(new Gson().toJson(userDTO.getAdditionalInfo()));
+        }
```
zhx828, line 365: I think this came from master branch?

src/main/java/org/mskcc/cbio/oncokb/web/rest/CreateAssessment.java
```diff
@@ -0,0 +1,72 @@
...
+   *                        recaptchaSiteKey.
+   * @param recaptchaAction : Action name corresponding to the token.
+   * @return
+   * @throws ValidationException
+   */
+  public static ResponseEntity<String> createAssessment(HttpServletRequest request, @RequestParam String recaptchaToken, String recaptchaAction)
+      throws IOException, ValidationException {
+ 
+    String projectID = "symbolic-nation-320615";
+    String recaptchaSiteKey = "6LdTXvMhAAAAAN7kj4MRKX0fl_gXUv_IQbxARe6W";
```
zhx828, line 38: These need to be specified through application properties

src/main/webapp/app/config/constants.tsx
```diff
@@ -24,6 +24,8 @@ export const DISABLE_BANNER_OPT = 'disablebanner';
 export const DEV_URL = 'http://localhost:9095';
 
 export const RECAPTCHA_SITE_KEY = '6LcxRsMZAAAAAFYpXX6KAc9ASGSf8IptsIKehJby';
+export const RECAPTCHA_ENTERPRISE_SITE_KEY =
```
zhx828, line 27: Do we need both?

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/config/SecurityConfiguration.java
```diff
@@ -69,7 +69,7 @@ public void configure(HttpSecurity http) throws Exception {
...
-            .contentSecurityPolicy("default-src 'self'; frame-src 'self' https://*.google.com https://recaptcha.net https://*.oncokb.org https://*.youtube.com https://*.bilibili.com https://*.gitbook.com; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://storage.googleapis.com www.google-analytics.com https://recaptcha.net https://www.gstatic.com; style-src 'self' 'unsafe-inline'; img-src 'self' data: www.google-analytics.com; font-src 'self' data:; connect-src 'self' https://*")
+            .contentSecurityPolicy("default-src 'self'; frame-src 'self' https://*.google.com https://www.recaptcha.net https://*.oncokb.org https://*.youtube.com https://*.bilibili.com https://*.gitbook.com; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://www.google.com/ https://storage.googleapis.com www.google-analytics.com https://recaptcha.net https://www.gstatic.com; style-src 'self' 'unsafe-inline'; img-src 'self' data: www.google-analytics.com; font-src 'self' data:; connect-src 'self' https://*; script-src-elem 'self' https://www.google.com/ https://storage.googleapis.com www.google-analytics.com https://www.recaptcha.net https://www.gstatic.com 'unsafe-inline' 'unsafe-eval'")
```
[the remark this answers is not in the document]
scarrero4660: I just tested script-src alone and it works fine so I will update this

src/main/java/org/mskcc/cbio/oncokb/web/rest/CreateAssessment.java
scarrero4660: This key is the public key while the private key and project ID is set in the environment variable GOOGLE_APPLICATION_CREDENTIALS. Is it still necessary to set the public key in the application properties? If so I suppose I can just add it to the GOOGLE_APPLICATION_CREDENTIALS

src/main/webapp/app/config/constants.tsx
scarrero4660: I believe I deleted the unnecessary key. Will double check

src/main/java/org/mskcc/cbio/oncokb/web/rest/CreateAssessment.java
zhx828: yes, but not to application properties. It should be added into application yaml files.

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -95,61 +95,85 @@ public AccountResource(UserRepository userRepository, UserService userService,
...
+                    throw new InvalidPasswordException();
+                }
+                User user = userService.registerUser(managedUserVM, managedUserVM.getPassword());
+                mailService.sendActivationEmail(userMapper.userToUserDTO(user));
+            }
+        } catch (Exception e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
+            String errorMessage = e.getMessage();
+            throw new Exception(errorMessage);
```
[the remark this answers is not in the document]
scarrero4660: If creating assessment failed then a ValidationException would be thrown and this message is displayed on the registration page

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -95,61 +95,85 @@ public AccountResource(UserRepository userRepository, UserService userService,
...
-                        token.setExpiration(expirationBased.isBefore(defaultExpiration) ? defaultExpiration : expirationBased);
+                        // if the extended date based on the current token expiration is before the date
+                        // in 6month, we should use the bigger one
+                        Instant expirationBased = token.getExpiration()
+                                .plusSeconds(tokenProvider.EXPIRATION_TIME_IN_SECONDS);
+                        token.setExpiration(
+                                expirationBased.isBefore(defaultExpiration) ? defaultExpiration : expirationBased);
                         tokenService.save(token);
                     });
                 }
```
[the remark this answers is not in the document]
scarrero4660: This might have been from VSCode auto-formatting

src/main/java/org/mskcc/cbio/oncokb/web/rest/CreateAssessment.java
```diff
@@ -0,0 +1,82 @@
...
+          .setParent(ProjectName.of(applicationProperties.getRecaptchaProjectId()).toString())
+          .setAssessment(Assessment.newBuilder().setEvent(event).build())
+          .build();
+
+      Assessment response = client.createAssessment(createAssessmentRequest);
+
+      client.close();
+
+      // Check if the token is valid.
+      if (response.getTokenProperties().getValid() && response.getRiskAnalysis().getScore() >= 0.5) {
```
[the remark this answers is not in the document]
scarrero4660: Yes I will add this

src/main/java/org/mskcc/cbio/oncokb/config/application/ApplicationProperties.java
```diff
@@ -35,6 +35,8 @@ public class ApplicationProperties {
     private String githubToken;
     private Boolean dbReadOnly;
     private SmartsheetProperties smartsheet;
+    private String recaptchaProjectId;
+    private String recaptchaSiteKey;
```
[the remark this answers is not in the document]
scarrero4660: Yes I will add that in my next commit

src/main/resources/config/application-dev.yml
```diff
@@ -169,3 +169,5 @@ application:
     s3-region:
   github-token:
   db-read-only: false # Certain endpoints will be disabled when readonly is set to true
+  recaptcha-project-id: 'symbolic-nation-320615'
+  recaptcha-site-key: '6LdTXvMhAAAAAN7kj4MRKX0fl_gXUv_IQbxARe6W'
```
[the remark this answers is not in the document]
scarrero4660: How would these values be set if they are not added here to be processed in the RecaptchaProperties class?

src/main/java/org/mskcc/cbio/oncokb/web/rest/CreateAssessment.java
```diff
@@ -0,0 +1,84 @@
...
+          .setParent(ProjectName.of(recaptchaProperties.getRecaptchaProjectId()).toString())
+          .setAssessment(Assessment.newBuilder().setEvent(event).build())
+          .build();
+
+      Assessment response = client.createAssessment(createAssessmentRequest);
+
+      client.close();
+
+      // Check if the token is valid.
+      if (response.getTokenProperties().getValid() && response.getRiskAnalysis().getScore() >= Long.valueOf(recaptchaProperties.getRecaptchaThreshold())) {
```
calvinlu3, line 68:
```suggestion
      if (response.getTokenProperties().getValid() && response.getRiskAnalysis().getScore() >= Float.valueOf(recaptchaProperties.getRecaptchaThreshold())) {
```
We should use `float` type here since Long is for whole numbers. The threshold can only be from 0.0 to 1.0 since thats what is being compared with the risk analysis score.

We need to update the properties file to change type to `Float`

src/main/resources/config/application-dev.yml
```diff
@@ -163,9 +163,16 @@ application:
...
   aws:
     s3-access-key:
     s3-secret-key:
     s3-region:
   github-token:
   db-read-only: false # Certain endpoints will be disabled when readonly is set to true
+  recaptcha:
+    recaptcha-project-id:
+    recaptcha-site-key:
+    recaptcha-threshold:
```
calvinlu3, line 178: @scarrero4660 @zhx828 Any reason we need to duplicate the properties here?

zhx828, line 178: Can we get rid of the prefix? We don't need the recaptcha if they already under the recaptcha property

scarrero4660: Yes I can get rid of that

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -322,39 +368,71 @@ public void deleteToken(@RequestBody Token token) throws AuthenticationException
...
-    public void requestPasswordReset(@RequestBody String mail) {
-        Optional<User> user = userService.getUserWithAuthoritiesByEmailIgnoreCase(mail);
-        if (user.isPresent()) {
-            Optional<User> updatedUser = userService.requestPasswordReset(user.get().getLogin());
-            if (updatedUser.isPresent()) {
-                mailService.sendPasswordResetMail(userMapper.userToUserDTO(updatedUser.get()));
+    public void requestPasswordReset(@RequestBody String mail, HttpServletRequest request) throws Exception {
+        try {
+            ResponseEntity<String> rs = CreateAssessment.createAssessment(request);
+            // if (rs.getStatusCode() == HttpStatus.OK) {
```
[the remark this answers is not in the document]
scarrero4660: This is temporarily commented out so that a poor score will not block the user from resetting their password. We can un-comment this after the trial period

src/main/webapp/app/shared/recaptcha/recaptcha.ts
```diff
@@ -0,0 +1,37 @@
...
+    // this.siteKey = RECAPTCHA_ENTERPRISE_SITE_KEY;
+  }
+
+  async getToken(): Promise<string> {
+    let token = '';
+    await window.grecaptcha.enterprise
+      .execute(this.siteKey)
+      .then((res: string) => {
+        token = res;
+      });
```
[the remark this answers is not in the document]
scarrero4660: Yes I am looking into it. Thanks for the suggestion

src/test/java/org/mskcc/cbio/oncokb/web/rest/AccountResourceIT.java
```diff
@@ -131,7 +131,8 @@ public void testRegisterValid() throws Exception {
         restAccountMockMvc.perform(
             post("/api/register")
                 .contentType(MediaType.APPLICATION_JSON)
-                .content(TestUtil.convertObjectToJsonBytes(validUser)))
+                .content(TestUtil.convertObjectToJsonBytes(validUser))
+                .header("g-recaptcha-response", "faketoken"))
```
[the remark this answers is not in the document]
scarrero4660: Okay I will change this

src/main/java/org/mskcc/cbio/oncokb/config/application/FrontendProperties.java
```diff
@@ -6,6 +6,33 @@
 public class FrontendProperties {
     private String googleAnalyticsProjectId;
     private String sentryProjectId;
+    String recaptchaProjectId;
+    String recaptchaSiteKey;
+    Long recaptchaThreshold;
```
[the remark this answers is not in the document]
scarrero4660: The only prop really used here is the recaptchaProjectId which is used in the frontend recaptcha class. It is sourced from the serverConfig which is set as the FrontendProperties in the ClientForwardController.java file

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -95,61 +96,92 @@ public AccountResource(UserRepository userRepository, UserService userService,
...
+                if (!checkPasswordLength(managedUserVM.getPassword())) {
+                    throw new InvalidPasswordException();
+                }
+                User user = userService.registerUser(managedUserVM, managedUserVM.getPassword());
+                mailService.sendActivationEmail(userMapper.userToUserDTO(user));
+            }
+        } catch (ValidationException e) {
+            e.printStackTrace();
+            String errorMessage = e.getMessage();
+            if (errorMessage.contains("Unable to retrieve recaptcha token.")) {
```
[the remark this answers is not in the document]
scarrero4660: Ok I will add this

scarrero4660:
You can access and generate testing keys and project keys on the [Recaptcha Enterprise page](https://console.cloud.google.com/security/recaptcha?_ga=2.130768143.1418812120.1675264090-363223125.1663089556&_gac=1.178441488.1674717413.EAIaIQobChMIvYjOwdjk_AIVlLjICh1fSgcVEAAYASAAEgLsVfD_BwE&pli=1&project=symbolic-nation-320615) on Google Cloud.

src/main/java/org/mskcc/cbio/oncokb/config/Constants.java
```diff
@@ -40,6 +40,10 @@ public final class Constants {
 
     public static final String ONCOKB_TM = "OncoKB™";
 
+    public static final String VALIDATION_ERROR = "Validation failed";
+    public static final String TESTING_TOKEN = "faketoken";
```
zhx828, line 44: These names need a prefix.

src/main/java/org/mskcc/cbio/oncokb/config/Constants.java
```diff
@@ -40,6 +40,10 @@ public final class Constants {
 
     public static final String ONCOKB_TM = "OncoKB™";
 
+    public static final String VALIDATION_ERROR = "Validation failed";
```
zhx828, line 43: I meant to add the variable as constant in CreateAssesment file since it's not shared with other classes.

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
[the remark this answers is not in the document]
zhx828: `Unable to retrieve recaptcha token.` needs to be a constant in this file.

src/main/webapp/app/shared/recaptcha/recaptcha.ts
```diff
@@ -0,0 +1,37 @@
+import { RECAPTCHA_ENTERPRISE_SITE_KEY } from 'app/config/constants';
```
zhx828, line 1: This is not available and not needed.

src/main/java/org/mskcc/cbio/oncokb/web/rest/CreateAssessment.java
```diff
@@ -0,0 +1,81 @@
...
+
+    String recaptchaToken = request.getHeader("g-recaptcha-response");
+    if (recaptchaToken == null) {
+      throw new ValidationException("Unable to retrieve recaptcha token. Please try again.");
+    }
+
+    try (RecaptchaEnterpriseServiceClient client = RecaptchaEnterpriseServiceClient.create()) {
+
+      // Set the properties of the event to be tracked.
+      Event event = Event.newBuilder().setSiteKey(recaptchaProperties.getSiteKey()).setToken(recaptchaToken).build();
```
zhx828, line 52: When recaptcha properties are not defined, we should print a log indicating recaptcha enterprise is not initiated.

src/main/webapp/app/shared/recaptcha/recaptcha.ts
```diff
@@ -0,0 +1,37 @@
...
+  const script = document.createElement('script');
+  script.src = `https://www.google.com/recaptcha/enterprise.js?render=${siteKey}`;
+  document.body.appendChild(script);
+};
+
+export default class ReCAPTCHA {
+  siteKey: string;
+
+  constructor() {
+    this.siteKey = AppConfig.serverConfig.recaptchaSiteKey;
```
zhx828, line 21: `recaptchaSiteKey` is not always available, we need to handle the case.

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/config/application/RecaptchaProperties.java
```diff
@@ -0,0 +1,32 @@
+package org.mskcc.cbio.oncokb.config.application;
+
+public class RecaptchaProperties {
+    String projectId;
+    String siteKey;
+    float threshold;
```
zhx828, line 6: All props should be private.

src/main/java/org/mskcc/cbio/oncokb/config/application/FrontendProperties.java
zhx828, line 11: They should be private

src/test/resources/config/application-testcontainers.yml
```diff
@@ -25,3 +25,22 @@ spring:
...
+
+application:
+  frontend:
+    recaptcha-project-id:
+    recaptcha-site-key:
+    recaptcha-threshold:
+  recaptcha:
+    project-id:
+    site-key:
+    threshold:
```
zhx828, line 46: Do we need to specify these?

src/test/java/org/mskcc/cbio/oncokb/web/rest/RecaptchaAssessmentIT.java
```diff
@@ -0,0 +1,153 @@
...
+    //             UUID.randomUUID().toString(),
+    //             Arrays.<MockGrpcService>asList(mockRecaptchaEnterpriseService));
+    //     mockServiceHelper.start();
+    // }
+
+    // @AfterAll
+    // public static void stopServer() {
+    //     mockServiceHelper.stop();
+    // }
+
```
zhx828, line 61: We should do some cleanup here.

src/test/java/org/mskcc/cbio/oncokb/web/rest/MockRecaptchaEnterpriseService.java
```diff
@@ -0,0 +1,58 @@
+// package org.mskcc.cbio.oncokb.web.rest;
```
zhx828, line 1: If we don't use these services, we should just delete.

src/test/java/org/mskcc/cbio/oncokb/web/rest/AccountResourceIT.java
```diff
@@ -58,6 +62,25 @@ public class AccountResourceIT {
...
+    @Autowired
+    private ApplicationProperties applicationProperties;
+
+    RecaptchaProperties recaptchaProp;
+
+    private AccountResource accountResource;
+
+    @BeforeEach
+    public void setUp() throws IOException {
+        // applicationProperties = new ApplicationProperties();
```
zhx828, line 74: clean up

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -408,14 +445,26 @@ public UserDTO getTrialAccountActivationInfo(@RequestParam String key) {
...
+        try {
+            RecaptchaEnterpriseServiceClient client = createAssess.createClient();
+            String recaptchaToken = createAssess.getRecaptchaToken(request);
+            ResponseEntity<String> rs = createAssess.createAssessment(client,recaptchaToken);
+        } catch (ValidationException e) {
+            e.printStackTrace();
+            String errorMessage = e.getMessage();
+            if (errorMessage.contains("Unable to retrieve recaptcha token.")) {
+                throw new Exception(errorMessage);
+            }
```
zhx828, line 459: Is there a way for us to not throw exception until we gather enough info whether token is attached on every request?

pom.xml
```diff
@@ -98,18 +98,128 @@
...
+
+        <!-- <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-stub</artifactId>
+        </dependency>
+        <dependency>
```
zhx828, line 136: Let's clean up all unused dependency

src/main/resources/config/application-dev.yml
```diff
@@ -163,9 +163,16 @@ application:
...
   aws:
     s3-access-key:
     s3-secret-key:
     s3-region:
   github-token:
   db-read-only: false # Certain endpoints will be disabled when readonly is set to true
+  recaptcha:
+    project-id:
+    site-key:
+    threshold:
```
zhx828, line 178: Are these required? Would application run without these defined here?

src/main/webapp/app/App.tsx
```diff
@@ -92,6 +96,13 @@ class App extends React.Component {
     return (
       <DocumentTitle title={DOCUMENT_TITLES.HOME}>
         <>
+          {/* <Reaptcha
+            ref={this.recaptchaRef}
+            sitekey={RECAPTCHA_SITE_KEY}
+            onVerify={this.onRecaptchaVerify}
+            onRender={this.onRecaptchaRender}
+            size="invisible"
+          /> */}
```
zhx828, line 105: Let's remove

zhx828 (changes requested):
We should remove all unused fragments(code, comment). If you want to reuse them in the future, you can create another branch to keep a copy.

src/test/resources/config/application-testcontainers.yml
scarrero4660: I believe this is unnecessary now that the `AccountResource` file is updated. Now removed

src/main/resources/config/application-dev.yml
scarrero4660: Yes these are required because the `RecaptchaProperties` are used in `CreateAssessment` to load the site key and project ID

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
scarrero4660: I believe the recaptcha enterprise console would only provide information about the scores from each request and not the specific error for why the request would fail but I can double check

calvinlu3:
@scarrero4660 
From a user standpoint, I tested the /register endpoint and it works well for me. 

I did notice that the recaptcha symbol at the bottom right corner of the page will remain even when I have navigated off of the /register page.
Steps to reproduce:
- Land on OncoKB news page (or any non-protected pages). The recaptcha symbol won't show up, as expected
- Navigate to the registration page. Here the recaptcha symbol will show, as expected
- Navigate back to the news page. The recaptcha icon will persist

src/main/webapp/app/shared/recaptcha/recaptcha.ts
```diff
@@ -0,0 +1,44 @@
...
+  async getToken(): Promise<string> {
+    let token = '';
+    if (this.siteKey !== '' && this.siteKey !== null) {
+      await window.grecaptcha.enterprise
+        .execute(this.siteKey)
+        .then((res: string) => {
+          token = res;
+        })
+        .catch((error: Error) => {
+          notifyError(error);
```
zhx828, line 39: We should not send error message to user but rather send to sentry. See https://github.com/oncokb/oncokb-public/blob/master/src/main/webapp/app/index.tsx#L99 for an example.

zhx828 (changes requested):

PR #908, opened 2022-12-05
Update token usage text

zhx828:
- use default US locale number
- use text-danger if number is bigger than 1B
- use text-warning if number is bigger than 1M

Screen Shot 2022-12-05 at 3 13 02 PM


calvinlu3 (approved):
Looks good

PR #911, opened 2022-12-06
Define defaultPageSize instead of size for most of OncoKBTable usage

zhx828:
Once the pagesize is fixed, the page size dropdown no longer works.


calvinlu3 (approved):
Looks good

PR #939, opened 2023-02-02
Allow enable/disable authentication in the frontend

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3373


calvinlu3 (approved):
We should probably also disable the login route in routes.tsx? The login page is still reachable by changing the url.

PR #942, opened 2023-02-06
Fix the relevant tumor type search keyword issue

zhx828:
The tumor type search key will be the code when subtype is selected, therefore the original string array relevantCancerTypes cannot identify the code.
This fixes https://github.com/oncokb/oncokb/issues/3390


zhx828:
@calvinlu3 don't review yet.

calvinlu3 (approved):

PR #956, opened 2023-03-21
Do not display lollipop in gene page if no portal alterations are related

zhx828:


calvinlu3:
This change would remove the bar chart as well?

zhx828:
> This change would remove the bar chart as well?

Good point, I didn't think of it initially. I took another look, it should remove the bar chart as well. Bar chart also relies on the same data.

calvinlu3:
LGTM

PR #959, opened 2023-03-29
Add navigation breadcrumbs for alteration and cancer type pages

zhx828:
### Alteration page
Screen Shot 2023-03-29 at 4 52 43 PM

### Cancer type page
Screen Shot 2023-03-29 at 4 52 36 PM

### DIFF
--- a/src/main/webapp/app/variables.scss
+++ b/src/main/webapp/app/variables.scss
@@ -16,6 +16,11 @@ $danger: #dc3545;
 $link-hover-color: $oncokb-darker-blue;
 $nav-bg-color: $oncokb-blue;
 
+$breadcrumb-bg: transparent;
+$breadcrumb-padding-x: 0;
+$breadcrumb-padding-y: 0;
+$breadcrumb-margin-bottom: 0.5rem;
+
 $navbar-padding-x: 0;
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/variables.scss
```diff
@@ -16,6 +16,11 @@ $danger: #dc3545;
 $link-hover-color: $oncokb-darker-blue;
 $nav-bg-color: $oncokb-blue;
 
+$breadcrumb-bg: transparent;
+$breadcrumb-padding-x: 0;
+$breadcrumb-padding-y: 0;
```
calvinlu3, line 21: Padding of `0` is default if you don't define them here?

zhx828: I can double check, I don't think they are 0.

zhx828:
@calvinlu3 I will update the screenshots when all features are done.

src/main/webapp/app/variables.scss
zhx828: This is the default.
```
$breadcrumb-padding-y:              .75rem !default;
$breadcrumb-padding-x:              1rem !default;
```

PR #960, opened 2023-03-30
Include drug description column and shorten the text by number of words

zhx828:
- Include drug description column and shorten the text by number of words
- Include cancer type in the page header

### logged in
Screen Shot 2023-03-30 at 11 01 55 AM
Screen Shot 2023-03-30 at 11 02 00 AM

### logged out
Screen Shot 2023-03-30 at 11 03 11 AM

### DIFF
--- a/src/main/webapp/app/oncokb-frontend-commons/src/components/LongText.tsx
+++ b/src/main/webapp/app/oncokb-frontend-commons/src/components/LongText.tsx
@@ -1,0 +1,36 @@
+import * as React from 'react';
+import { useState } from 'react';
+import SummaryWithRefs from './SummaryWithRefs';
+
+const shortenTextByWords = (text: string, numOfWords: number) => {
+  const separator = ' ';
+  return text.split(separator).slice(0, numOfWords).join(separator);
+};
+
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/oncokb-frontend-commons/src/components/LongText.tsx
```diff
@@ -0,0 +1,36 @@
...
+      {expandedText || text.length === props.text.length ? undefined : (
+        <>
+          <span className={'mx-2'}>...</span>
+          <span
+            className={'text-primary'}
+            style={{ cursor: 'pointer' }}
+            onClick={() => setExpandedText(true)}
+          >
+            Show more
+          </span>
```
calvinlu3, line 31: We're getting rid of the underline on `Show more`?

calvinlu3 (approved):
LGTM.

src/main/webapp/app/oncokb-frontend-commons/src/components/LongText.tsx
zhx828: Good point, I think it should still add it. I will add in next pr.

PR #965, opened 2023-04-04
Minor changes to cancer type page based on feedback

zhx828:


src/main/webapp/app/oncokb-frontend-commons/src/components/LongText.tsx
```diff
@@ -2,27 +2,35 @@ import * as React from 'react';
 import { useState } from 'react';
 import SummaryWithRefs from './SummaryWithRefs';
 import styles from '../../../index.module.scss';
+import classnames from 'classnames';
+import showHideText from 'app/shared/texts/ShowHideText';
```
calvinlu3, line 6: Not used

src/main/webapp/app/oncokb-frontend-commons/src/components/LongText.tsx
```diff
@@ -2,27 +2,35 @@ import * as React from 'react';
...
 
 const shortenTextByCharacters = (text: string, cutoff: number) => {
```
calvinlu3, line 8: Seems like this can be reused elsewhere. Maybe move to a utils file?

calvinlu3 (approved):
LGTM
Do we also want to make the book icon in description column to have the style cursor: pointer?

zhx828:
> LGTM Do we also want to make the book icon in description column to have the style cursor: pointer?

@calvinlu3 it's not clickable, I would not.

PR #966, opened 2023-04-05
Fix: should show warning message when gene not exists

zhx828:


src/main/webapp/app/store/AnnotationStore.ts
```diff
@@ -291,20 +292,15 @@ export class AnnotationStore {
...
-      highestPrognosticImplicationLevel: '',
-      highestFdaLevel: '',
-      tumorType: 0,
+      try {
+        return await privateClient.utilsNumbersGeneGetUsingGET({
+          hugoSymbol: this.gene.result.hugoSymbol,
+        });
+      } catch (e) {
+        return DEFAULT_GENE_NUMBER;
+      }
```
calvinlu3, line 301: Another option beside `try...catch` is to use `onError: () => { return DEFAULT_GENE_NUMBER }` ?

PR #968, opened 2023-04-05
Make cancer type page mobile friendly

zhx828:
- Only show description column when row has description
- Include bootstrap table-responsive in the SimpleTable component
- Update cancer type page view when browser width smaller than large grid

### view when in mobile
Screen Shot 2023-04-05 at 3 05 56 PM


src/main/webapp/app/pages/annotationPage/CancerTypeViewTable.tsx
```diff
@@ -27,35 +30,50 @@ export interface ITableRow {
...
-        Associated with{' '}
-        {altNameIncludesHugoSymbol ? '' : `${props.hugoSymbol} `}
-        {props.alterations}
-      </div>
-      <div className={'text-secondary'}>
-        in {props.cancerTypes.toLowerCase()}
+        {geDrugAssociationCellText(
+          props.hugoSymbol,
+          props.alterations,
+          props.cancerTypes
```
calvinlu3, line 61: No longer converting to lowercase?

src/main/webapp/app/pages/annotationPage/CancerTypeViewTable.tsx
```diff
@@ -111,8 +134,114 @@ function getRows(
...
+      },
+    ];
+    return {
+      key: `${type}-row-${index}`,
+      content,
+    };
+  });
+}
+
+function getNoneCompactTxRows(
```
calvinlu3, line 224:
```suggestion
function getNonCompactTxRows(
```

calvinlu3 (approved):

src/main/webapp/app/pages/annotationPage/CancerTypeViewTable.tsx
zhx828: Yeah, some cancer type names include genetic biomarkers. They are incorrect to be converted to lowercase. I'm thinking to make it simple without any conversion, similar to what we show on the title. If later user requests the name to be converted to lowercase, then we will spend more time fixing.

PR #973, opened 2023-04-28
Handel redis caching errors

zhx828:
We will print all errors in log but let it silently skip so website functionality is not affected.

This is related to https://github.com/oncokb/oncokb/issues/3326


src/main/java/org/mskcc/cbio/oncokb/config/cache/CachingConfiguration.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class CachingConfiguration extends CachingConfigurerSupport {
+    @Override
+    public CacheErrorHandler errorHandler() {
+        return new LoggingCacheErrorHandler();
+    }
+}
+
```
calvinlu3, line 14: I'm wondering if its the same if we extend CacheConfiguration (`src/main/java/org/mskcc/cbio/oncokb/config/CacheConfiguration.java`) with `CachingConfigurerSupport` and then register the errorHandler in the same file?

zhx828: That's a good point. Let me try.

zhx828: @calvinlu3 updated

calvinlu3 (approved):
LGTM

PR #974, opened 2023-04-28
Update debian imported stretch suite location

zhx828:
The location of deb.debian has been moved to archive. We need to change the source without modifying the node image


calvinlu3 (approved):
LGTM, building the image works for me now.  👍

PR #977, opened 2023-05-01
Allow admin to download company user list

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/108

### Changes

1. Added download button 
image

Preview of downloaded tsv file
image

2. Moved the download icon into `DownloadButton` component
3. Remove `Button` component and just made the `a` tag use button classnames. In the original implementation, the download is triggered only when you click on the text inside the anchor tag and not the button itself. For instance https://www.oncokb.org/levels, download the pdf doesn't work if you don't click on the text exactly.


calvinlu3:
@zhx828 Are there any other user data fields that we want to download besides email and name?

zhx828 (approved):

PR #981, opened 2023-06-01
Slack email abstraction

benzuzu:
make adding new email options in slack easier by encapsulating all related data into the MailType enum.


src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
```diff
@@ -1,45 +1,196 @@
...
 
-    String templateName;
-    String description;
-    String attachmentFileNames;
 
-    MailType(String templateName, String description, String attachmentFileNames) {
-        this.templateName = templateName;
-        this.description = description;
-        this.attachmentFileNames = attachmentFileNames;
+    private static class MailTypeBuilder {
```
benzuzu, line 65: used builder pattern since there are so many properties, can assign each one rather than use a single constructor with some arbitrary order

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
```diff
@@ -1,45 +1,196 @@
...
-    , LIST_OF_UNAPPROVED_USERS("listOfUnapprovedUsers", "List of unapproved users", null)
-    , DATA_USAGE_EXCEEDS_THRESHOLD("dataUsageExceedsThreshold", "Unusual OncoKB account activities (data usage)", null)
-    , TEST("testEmail", "Test", null)
+    ACTIVATION(new MailTypeBuilder()
+        .templateName("activationEmail")
+        .description("User Activation"))
+    , APPROVAL(new MailTypeBuilder()
+        .templateName("approvalEmail")
+        .description("User Approval"))
+    , REJECTION(new MailTypeBuilder()) //"rejectionEmail", "User Rejection", null)
```
benzuzu, line 18: only filled in a few of these, will complete the rest if you think this approach is ok.

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
```diff
@@ -136,31 +136,9 @@ public ResponseEntity<String> approveUser(@RequestParam("payload") String action
...
-                    case CONFIRM_SEND_LICENSE_OPTIONS_EMAIL:
-                        mailType = MailType.LICENSE_OPTIONS;
-                        break;
-                    case CONFIRM_SEND_REJECTION_EMAIL:
-                        mailType = MailType.REJECTION;
-                        break;
-                    case CONFIRM_SEND_REJECT_ALUMNI_ADDRESS_EMAIL:
-                        mailType = MailType.REJECT_ALUMNI_ADDRESS;
-                        break;
+                for (MailType curMailType : MailType.values()) {
```
benzuzu, line 139: can now iterate through the enum without having to do each case

src/main/java/org/mskcc/cbio/oncokb/web/rest/slack/ActionId.java
```diff
@@ -56,45 +58,39 @@ public static boolean isEmailAction(ActionId actionId) {
...
-            || actionId == COLLAPSE;
+
+        // mail actions
+        for (MailType mailType : MailType.values()) {
+            if (actionId.equals(mailType.getActionId().orElse(null)))
+                return true;
+        }
+
+        // non mail actions
+        if (actionId == CONVERT_TO_REGULAR_ACCOUNT || actionId == UPDATE_USER || actionId == COLLAPSE)
```
benzuzu, line 91: some dropdown actions do not have emails associated with them, so if we wish to add more of those, the logic will still have to be modified in several different classes.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -19,6 +19,8 @@
 import com.slack.api.model.view.ViewTitle;
 import com.slack.api.webhook.Payload;
 import com.slack.api.webhook.WebhookResponse;
+import com.smartsheet.api.models.Email;
+import liquibase.pro.packaged.U;
```
benzuzu, line 23: these are unnecessary, will remove later

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -399,85 +398,54 @@ public String getOptionValueArgument(String value) {
         return values[0];
     }
 
-    public boolean withTrialAccountNote(UserDTO userDTO, ActionId actionId) {
-        if (
-            ObjectUtil.isObjectEmpty(userDTO.getAdditionalInfo())
-                || userDTO.getAdditionalInfo().getTrialAccount() == null
-                || userDTO.getAdditionalInfo().getTrialAccount().getActivation() == null
-        ) {
+    public boolean withNote(MailType mailType, UserDTO userDTO, ActionId actionId) {
```
benzuzu, line 401: created a general withNote function that checks if a certain MailType has been sent/was just sent. this works for the regular emails we wish to send (anything through the slack view modal). it will not work for automated ones as those have their own logic. so if we plan to add more automated emails the logic will still need to be changed here.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -553,36 +521,23 @@ private List<LayoutBlock> buildUserInfoBlocks(UserDTO user) {
...
-        }
-        if (withDuplicateUserClarificationNote(userDTO, actionId)) {
-            layoutBlocks.add(buildPlainTextBlock("We have sent a clarification email to the user asking why they registered multiple accounts.", DUPLICATE_USER_CLARIFICATION_NOTE));
-        }
-        if (withRegistrationInfoClarificationNote(userDTO, actionId)) {
-            layoutBlocks.add(buildPlainTextBlock("We have sent a clarification email to the user asking to provide detailed registration info.", REGISTRATION_INFO_CLARIFICATION_NOTE));
-        }
-        if (withLicenseOptionsNote(userDTO, actionId)) {
-            layoutBlocks.add(buildPlainTextBlock("We have sent an email to the user with license options for their affiliated company.", LICENSE_OPTIONS_NOTE));
+        for (MailType inquiryMail : Arrays.stream(MailType.values()).filter(mt -> mt.getCategory().orElse(null) == EmailCategory.CLARIFY || mt.getCategory().orElse(null) == EmailCategory.LICENSE).collect(Collectors.toList())) {
```
benzuzu, line 524: this is how we iterate through enum, can filter by clarification/license emails like here.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -263,24 +265,21 @@ public List<LayoutBlock> buildBlocks(UserDTO userDTO, boolean trialAccountActiva
...
-                sb.append("Sent license options email");
+        sb.append(userDTO.getEmail() + "\n" + userDTO.getCompanyName() + " (" + userDTO.getLicenseType().getShortName() + (withNote(MailType.ACTIVATE_FREE_TRIAL, userDTO, actionId) ? ", *TRIAL*)" : userDTO.isActivated() ? ")" : ")\n*NOT ACTIVATED*: "));
+        if (!userDTO.isActivated() && !withNote(MailType.ACTIVATE_FREE_TRIAL, userDTO, actionId)) {
+            List<MailType> sentMails = new ArrayList<>();
+            for (MailType mailType : MailType.values()) {
+                if (withNote(mailType, userDTO, actionId))
+                    sentMails.add(mailType);
+            }
+            if (!sentMails.isEmpty()) {
+                sentMails = sentMails.stream().sorted((mt1, mt2) -> mt1.getCategory().orElse(null) == EmailCategory.DENY ? 1 : 0).collect(Collectors.toList());
```
benzuzu, line 276: sort the rejection emails at the end, clarification/license emails before to have some sort of chronological nature still in the collapsed version.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -261,9 +261,11 @@ public List<LayoutBlock> buildBlocks(UserDTO userDTO, boolean trialAccountActiva
         return blocks;
     }
 
-    private LayoutBlock buildCollapsedBlock(UserDTO userDTO, ActionId actionId) {
+    private LayoutBlock buildCollapsedBlock(UserDTO userDTO, boolean trialAccountActivated, ActionId actionId) {
         StringBuilder sb = new StringBuilder();
-        sb.append(userDTO.getEmail() + "\n" + userDTO.getCompanyName() + " (" + userDTO.getLicenseType().getShortName() + (withNote(DropdownEmailOption.GIVE_TRIAL_ACCESS, userDTO, actionId) ? ", *TRIAL*)" : userDTO.isActivated() ? ")" : ")\n*NOT ACTIVATED*: "));
+        sb.append(userDTO.getEmail() + "\n" + userDTO.getCompanyName() + " (" + userDTO.getLicenseType().getShortName()
+            + ((withNote(DropdownEmailOption.GIVE_TRIAL_ACCESS, userDTO, actionId) && !(userDTO.isActivated() && !trialAccountActivated)) ? ", *TRIAL*)" :
```
benzuzu, line 267: this `(withNote(DropdownEmailOption.GIVE_TRIAL_ACCESS, userDTO, actionId) && !(userDTO.isActivated() && !trialAccountActivated))` is basically as follows: withNote is if we ever gave them trial access, and second boolean is the negation of conversion to regular account (not converted to regular yet). so the trial note will dissapear if we convert to regular.

benzuzu: this logic is used a few other places too

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -576,43 +578,34 @@ private ButtonElement buildApproveButton(UserDTO user) {
...
-            for (DropdownEmailOption denyOption : Arrays.stream(DropdownEmailOption.values()).filter(mo -> mo.getCategory() == EmailCategory.DENY).collect(Collectors.toList())) {
-                denyGroup.add(buildEmailOption(denyOption, user));
+            for (EmailCategory emailCategory : Arrays.stream(EmailCategory.values()).filter(ec -> !ec.equals(EmailCategory.TRIAL)).collect(Collectors.toList())) {
+                List<OptionObject> optionGroup = new ArrayList<>();
+                for (DropdownEmailOption emailOption : Arrays.stream(DropdownEmailOption.values()).filter(eo -> eo.getCategory().equals(emailCategory)).collect(Collectors.toList())) {
+                    if (emailOption.getSpecificLicenses().isEmpty() || emailOption.getSpecificLicenses().contains(user.getLicenseType()))
+                        optionGroup.add(buildEmailOption(emailOption, user));
+                }
+                if (!optionGroup.isEmpty()) {
+                    optionGroups.add(OptionGroupObject.builder().label(PlainTextObject.builder().text(emailCategory.getLabel()).build()).options(optionGroup).build());
```
benzuzu, line 606: by giving EmailCategory a label property we can generalize option group creation since each dropdown option has an EmailCategory. trial has its own because there is different logic with conversion to regular and giving trial access

src/main/java/org/mskcc/cbio/oncokb/web/rest/slack/DropdownEmailOption.java
```diff
@@ -149,6 +159,11 @@ public DropdownEmailOptionBuilder mailType(MailType mailType) {
             return this;
         }
 
+        public DropdownEmailOptionBuilder isNotModalEmail() {
```
benzuzu, line 162: some emails (only trial atm) do not require view modal, it used to throw an error as it expected view modal parameter but new logic is fine.

src/main/java/org/mskcc/cbio/oncokb/web/rest/slack/DropdownEmailOption.java
```diff
@@ -159,6 +174,11 @@ public DropdownEmailOptionBuilder category(EmailCategory category) {
             return this;
         }
 
+        public DropdownEmailOptionBuilder specificLicense(LicenseType licenseType) {
```
benzuzu, line 177: some emails should only show up for certain license type. i.e. domain clarification only for academic, license options only for commercial, research in commercial, hospital

src/main/java/org/mskcc/cbio/oncokb/web/rest/slack/DropdownEmailOption.java
```diff
@@ -0,0 +1,275 @@
...
+        .category(EmailCategory.DENY)
+        .dropdownKey("Send Alumni Rejection Email")
+        .modalTitle("Reject Alumni Address")
+        .modalSubject(EmailSubject.DEFAULT)
+        .collapsedNote("Rejected user due to alumni email address")
+        .expandedNote("The user has been rejected due to alumni email address."))
+    ;
+
+
+    private static class DropdownEmailOptionBuilder {
```
zhx828, line 123: Can we add comments on which fields are required? Probably even better to include the required fields(hopefully not that many) in the contractor.

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/EmailCategory.java
```diff
@@ -0,0 +1,19 @@
+package org.mskcc.cbio.oncokb.domain.enumeration;
+
+/**
+ * The EmailCategory enumeration.
+ */
+public enum EmailCategory {
```
zhx828, line 6: We should explain how these categories are differentiated.

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/EmailSubject.java
```diff
@@ -0,0 +1,9 @@
+package org.mskcc.cbio.oncokb.domain.enumeration;
+
+/**
+ * The EmailSubject enumeration.
+ */
+public enum EmailSubject {
```
zhx828, line 6: Here as well. We need details on these categories.

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
```diff
@@ -1,45 +1,164 @@
...
-    , TOKEN_HAS_BEEN_EXPOSED("tokenHasBeenExposed", "Token has been exposed", null)
-    , TOKEN_HAS_BEEN_EXPOSED_USER("tokenHasBeenExposedToUser", "Token has been exposed", null)
-    , SEARCHING_RESPONSE_STRUCTURE_HAS_CHANGED("searchingResponseStructureHasChanged", "Searching Response Structure Has Changed", null)
-    , LIST_OF_UNAPPROVED_USERS("listOfUnapprovedUsers", "List of unapproved users", null)
-    , DATA_USAGE_EXCEEDS_THRESHOLD("dataUsageExceedsThreshold", "Unusual OncoKB account activities (data usage)", null)
-    , TEST("testEmail", "Test", null)
+    ACTIVATION(new MailTypeBuilder()
+        .templateName("activationEmail")
+        .description("User Activation")
+        .titleKey("email.activation.title"))
```
zhx828, line 12: Typically we can add validation on class members. Could you look into whether enum class will obey the same rule? https://www.baeldung.com/spring-boot-bean-validation Then we can make these three required?

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
```diff
@@ -289,39 +287,6 @@ public void sendAcademicClarificationEmail(UserDTO user) {
...
-
-    @Async
-    public void sendRejectionEmail(UserDTO user) {
-        sendEmailWithLicenseContext(user, MailType.REJECTION, applicationProperties.getEmailAddresses().getRegistrationAddress(), null, null);
-    }
-
-    @Async
-    public void sendRejectAlumniAddressEmail(UserDTO user) {
-        sendEmailWithLicenseContext(user, MailType.REJECT_ALUMNI_ADDRESS, applicationProperties.getEmailAddresses().getRegistrationAddress(), null, null);
-    }
```
zhx828, line 323: Are we not using these methods anymore? Don't see the usage of them.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -261,26 +261,25 @@ public List<LayoutBlock> buildBlocks(UserDTO userDTO, boolean trialAccountActiva
...
+            + ((withNote(DropdownEmailOption.GIVE_TRIAL_ACCESS, userDTO, actionId) && !(userDTO.isActivated() && !trialAccountActivated)) ? ", *TRIAL*)" :
+            (userDTO.isActivated() ? ")" : ")\n*NOT ACTIVATED*: ")));
+        if (!userDTO.isActivated() && !withNote(DropdownEmailOption.GIVE_TRIAL_ACCESS, userDTO, actionId)) {
+            List<DropdownEmailOption> sentMails = new ArrayList<>();
+            for (DropdownEmailOption mailOption : DropdownEmailOption.values()) {
+                if (withNote(mailOption, userDTO, actionId))
+                    sentMails.add(mailOption);
+            }
+            if (!sentMails.isEmpty()) {
+                sentMails = sentMails.stream().sorted((mo1, mo2) -> mo1.getCategory() == EmailCategory.DENY ? 1 : 0).collect(Collectors.toList());
```
zhx828, line 276: can we just do 
```suggestion
                sentMails = sentMails.stream().sorted((mo1, mo2) -> mo1.getCategory() == EmailCategory.DENY).collect(Collectors.toList());
```

zhx828 (changes requested):
I will need some clarifications on how these categories in Enum classes are different from each other. I will do another review one you provide that.

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/EmailSubject.java
benzuzu: Right, will add header comments to both of these.

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
benzuzu: I believe `sorted()` takes an overridden `compare()` predicate to sort, which should return a number (negative means less than, zero means equal, positive means greater than). I wanted clarification and license emails to show up first and then rejection email last to have some sort of chronological order, so I made rejection email greater than all others. I don't think boolean `true` will automatically be converted to positive number.

I just realized that license options email should be after clarification emails as well, that would make more sense. Will add that to the changes.

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
benzuzu: Thanks for this suggestion, I looked into it and the static builder class can use the validation framework as well. I will go ahead and add the @NonNull constraint to the required fields. As for adding them to the constructor, there are 5 fields that are required. Would you want me to remove the builder pattern for those fields and require them in the builder constructor? @zhx828 This would make new option creation a bit more arbitrary (like before without builder, just matching parameter order instead of actually assigning) but I can also see how having so many fields with builder can also be confusing.

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
zhx828: Thank you for catching this.

src/main/java/org/mskcc/cbio/oncokb/web/rest/slack/ActionId.java
```diff
@@ -65,17 +65,6 @@ public static boolean isModalEmailAction(ActionId actionId) {
         return false;
     }
 
-    public static boolean isConfirmEmailAction(ActionId actionId) {
```
benzuzu, line 69: never used

src/main/java/org/mskcc/cbio/oncokb/web/rest/slack/DropdownEmailOption.java
```diff
@@ -10,10 +10,63 @@
...
+     .confirmActionId(ActionId)
+     .mailType(MailType)
+     .category(EmailCategory)
+     .specificLicense(LicenseType)
+     .dropdownKey("")
+     .modalTitle("")
+     .modalSubject(EmailSubject)
+     .collapsedNote("")
+     .expandedNote(""))
+ *
```
benzuzu, line 69: As it turns out, the bean validation does not actually work for enums. The bean validation only checks if a new instance of a class does not satisfy certain requirements during runtime, but all of these elements are enum constants determined at compile time. Instead, I decided to create some documentation to make it less likely for a new option to be created incorrectly. If you would prefer adding the 6 required properties as constructor parameters, I can do that as well.

src/main/java/org/mskcc/cbio/oncokb/web/rest/slack/ActionId.java
zhx828: TBH, I think we will be in a better place if creating a DropdownOption instead of DropdownEmailOption. Every option will have category, action id, email etc. But it's out of scope of this pr. I will create a separate issue to follow up.

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
```diff
@@ -1,45 +1,164 @@
...
-    , TRIAL_ACCOUNT_IS_ABOUT_TO_EXPIRE("trialAccountIsAboutToExpire", "Trail account is about to expire", null)
-    , TRIAL_ACCOUNT_IS_ACTIVATED("trialAccountIsActivated", "Trail Account is Activated", null)
-    , ACTIVATE_FREE_TRIAL("activateFreeTrial", "OncoKB Trial Activation Link", null)
-    , TOKEN_HAS_BEEN_EXPOSED("tokenHasBeenExposed", "Token has been exposed", null)
-    , TOKEN_HAS_BEEN_EXPOSED_USER("tokenHasBeenExposedToUser", "Token has been exposed", null)
-    , SEARCHING_RESPONSE_STRUCTURE_HAS_CHANGED("searchingResponseStructureHasChanged", "Searching Response Structure Has Changed", null)
-    , LIST_OF_UNAPPROVED_USERS("listOfUnapprovedUsers", "List of unapproved users", null)
-    , DATA_USAGE_EXCEEDS_THRESHOLD("dataUsageExceedsThreshold", "Unusual OncoKB account activities (data usage)", null)
-    , TEST("testEmail", "Test", null)
+    ACTIVATION(new MailTypeBuilder()
```
zhx828, line 9: Do we have no email with attachment file?

zhx828 (changes requested):
- We should move EmailCategory and Email Subject under the slack folder.
- Is it possible to add some tests?

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
benzuzu: Yes, that field has never been used. Not sure what the original purpose was.

src/test/java/org/mskcc/cbio/oncokb/service/SlackServiceIT.java
```diff
@@ -0,0 +1,130 @@
...
+
+        // Specify webhook address for testing
+        applicationProperties.setSlack(new SlackProperties());
+        applicationProperties.getSlack().setUserRegistrationWebhook(USER_REGISTRATION_WEBHOOK);
+
+        slackService = new SlackService(applicationProperties, mailService, emailService, userMailsService, userMapper, slack);
+    }
+
+    @Test
+    void testSendUserRegistrationToChannel() throws IOException {
```
benzuzu, line 78: This test is basically just making sure the payload we initially send to Slack is correct (contains the crucial information). Similar to MailServiceIT.testSendEmail, etc.

src/test/java/org/mskcc/cbio/oncokb/web/rest/SlackControllerIT.java
```diff
@@ -0,0 +1,257 @@
...
+        applicationProperties.getEmailAddresses().setTechDevAddress(TECH_DEV_ADDR);
+        jHipsterProperties.getMail().setFrom(DEFAULT_ADDR);
+
+        mailService = new MailService(jHipsterProperties, javaMailSender, messageSource, templateEngine, userMailsService, applicationProperties);
+        slackService = new SlackService(applicationProperties, mailService, emailService, userMailsService, userMapper, slack);
+        slackController = new SlackController(userService, userRepository, mailService, slackService, userMapper);
+    }
+
+    @Test
+    void testApproveUser() throws IOException, MessagingException {
```
benzuzu, line 146: This test mocks an approval payload for a mock user. Tests email integration, Slack integration, User activation, User Token, and UserMails. This is the main thing to be reviewed. If this type of test is what you are looking for I can make it for all the other cases.

src/test/java/org/mskcc/cbio/oncokb/web/rest/SlackControllerIT.java
```diff
@@ -0,0 +1,647 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.in;
+import static org.mockito.Mockito.*;
+import static org.mskcc.cbio.oncokb.config.Constants.DEFAULT_TOKEN_EXPIRATION_IN_SECONDS;
+
+@SpringBootTest(classes = OncokbPublicApp.class)
+public class SlackControllerIT {
+
+    // Testing Slack url and email addresses for applicationProperties
+    private static final String USER_REGISTRATION_WEBHOOK =  "https://hooks.slack.com/example";
```
zhx828, line 76: I see this URL been used multiple times. Is this what slack provided? Or it can be anything

src/test/java/org/mskcc/cbio/oncokb/web/rest/SlackControllerIT.java
```diff
@@ -0,0 +1,647 @@
...
+    private PasswordEncoder passwordEncoder;
+
+    private SlackController slackController;
+
+    @BeforeEach
+    void setUp() throws IOException, SlackApiException {
+
+        /******************************
+         * Set up app properties
+         ******************************/
```
zhx828, line 158: How is the log formatted?

src/test/java/org/mskcc/cbio/oncokb/web/rest/SlackControllerIT.java
```diff
@@ -0,0 +1,647 @@
...
+        // Add required values
+        Set<String> expectedValues = new HashSet<>();
+        expectedValues.add(DEFAULT_USER_EMAIL);
+        expectedValues.add(DEFAULT_COMPANY_NAME);
+        if (collapsed) {
+            expectedValues.add(BlockId.COLLAPSED.getId());
+            expectedValues.add(actionId == ActionId.CHANGE_LICENSE_TYPE ? OTHER_LICENSE_TYPE.getShortName() : DEFAULT_LICENSE_TYPE.getShortName());
+            if (actionId == ActionId.GIVE_TRIAL_ACCESS) {
+                expectedValues.add("TRIAL");
+            } else if (ActionId.isConfirmEmailAction(actionId)) {
```
zhx828, line 549: Curious why we don't add the collapsed noted as long as the DropdownEmailOption has.

zhx828: Is it to avoid unnecessary for loop? The same goes to line 558-563

src/test/java/org/mskcc/cbio/oncokb/web/rest/SlackControllerIT.java
```diff
@@ -0,0 +1,647 @@
...
+                if (sectionBlock.getFields() != null) {
+                    for (TextObject textObject : sectionBlock.getFields()) {
+                        expectedValues.removeIf(textObject.getText()::contains);
+                    }
+                }
+                if (sectionBlock.getBlockId() != null) {
+                    expectedValues.removeIf(sectionBlock.getBlockId()::contains);
+                }
+            }
+        }
```
zhx828, line 587: I see few places using this strategy. Is this the only option to test? What happens when we have same value in different positions within the block?

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
```diff
@@ -1,45 +1,168 @@
...
+        .templateName("activationEmail")
+        .description("User Activation")
+        .titleKey("email.activation.title"))
+    , CREATION(new MailTypeBuilder()
+        .templateName("creationEmail")
```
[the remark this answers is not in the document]
benzuzu: I was considering writing a script, but there are not too many so it was easier just to modify the format manually. The tests should probably deal with any typos in the file names, but I will double check now that you mention it.

src/test/resources/templates/mail/activateFreeTrialString.txt
```diff
@@ -0,0 +1 @@
+As part of your company's trial and evaluation of OncoKB, below is your activation link for a 3 month trial of OncoKB:
```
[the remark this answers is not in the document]
benzuzu: approvalEmailString is purely for testing. Didn't really know a better way to test the output and that the html file was being parsed correctly, so I went with this. It does pose a bit of an inconvenience though since changing the html file would require this test to be updated as well, so there are probably better solutions. What do you think?

src/test/java/org/mskcc/cbio/oncokb/web/rest/SlackControllerIT.java
benzuzu: It can be anything, is just a test string to ensure the same webhook makes it through to the end of the Slack integration.

src/test/java/org/mskcc/cbio/oncokb/web/rest/SlackControllerIT.java
benzuzu: Yes, it's mainly just so the cases that aren't dropdown emails don't have to go through the for loop. Definitely can omit and just add based on match though, is that better?

src/test/java/org/mskcc/cbio/oncokb/web/rest/SlackControllerIT.java
benzuzu: I wanted to make the test as general as possible, so just check every location of every block. Because we may change the format of the Slack Integration based on preferences that come up. This just ensures all the crucial information is at least somewhere in the blocks. If a value appears twice, it will pass on the first one and ignore the second. Thoughts?

src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/EmailSubject.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * This enum is for the Slack integration. When team member
+ * selects an email option, an input modal will pop up with
+ * one of the below options as a default subject. See
+ *
+ * SlackService.buildModalView()
+ *
+ * for more details.
+ */
+public enum EmailSubject {
```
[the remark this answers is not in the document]
benzuzu: Good idea, same goes for EmailCategory I think

src/test/resources/templates/mail/activateFreeTrialString.txt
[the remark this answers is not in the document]
benzuzu: Look into why the test methods are still calling from production directories, should call from test directories.

src/test/java/org/mskcc/cbio/oncokb/web/rest/SlackControllerIT.java
benzuzu: Search by BlockId (create issue)

src/test/resources/templates/mail/activateFreeTrialString.txt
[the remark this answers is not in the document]
benzuzu: Turns out the test methods were not calling from the production directories, the dropdown email files were just not found and the test was checking for an empty string, which passed. In the changes I have just passed a test string for subject/body and ensured the corresponding email has the same fields. The HTML parsing is actually tested in `MailServiceIT` anyway, so no need to test in the Slack tests.

benzuzu:
This also now fixes https://github.com/oncokb/oncokb/issues/2992

zhx828 (approved):

zhx828:
Looks good, thank you!

PR #993, opened 2023-07-05
initial display for cancer type 

Eric-Jiang1:
I added in the frontend logic for searching by cancer type.

Right now, it just displays the cancer name and links to the associated gene page

### DIFF
--- a/src/main/resources/config/application-dev.yml
+++ b/src/main/resources/config/application-dev.yml
@@ -37,7 +37,8 @@ spring:
       indent-output: true
   datasource:
     type: com.zaxxer.hikari.HikariDataSource
-    url: jdbc:mysql://localhost:3306/oncokb_public?useUnicode=true&characterEncoding=utf8&useSSL=false&useLegacyDatetimeCode=false&serverTimezone=UTC&createDatabaseIfNotExist=true
+    # changed database name
+    url: jdbc:mysql://localhost:3306/oncokb-public?useUnicode=true&characterEncoding=utf8&useSSL=false&useLegacyDatetimeCode=false&serverTimezone=UTC&createDatabaseIfNotExist=true
     username: root
     password: root
     hikari:
@@ -134,7 +135,7 @@ sentry:
 application:
   name: local
   profile: dev
-  api-proxy-url: 'http://localhost:8080/web_war'
+  api-proxy-url: 'http://localhost:8080/oncokb'
   sitemap-enabled: false
   google-webmaster-verification:
   token-usage-check:
@@ -149,7 +150,8 @@ application:
     user-registration-channel-id:
   redis:
     # Add the following if you want to opt in to using Redis
-    # enabled: true
+    # Not using Redis
+    enabled: false
     type: 'single'
     password: 'oncokb-public-redis-password'
     address: 'redis://localhost:6379'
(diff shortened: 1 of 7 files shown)


src/main/webapp/app/pages/HomePage.tsx
```diff
@@ -46,6 +46,7 @@ interface IHomeProps {
 export type ExtendedTypeaheadSearchResp = TypeaheadSearchResp & {
   alterationsName: string;
   tumorTypesName: string;
+  cancerName: string;
```
zhx828, line 49: Can we reuse tumorTypesName?

src/main/webapp/app/shared/utils/Utils.tsx
```diff
@@ -157,6 +157,10 @@ export function getAllTumorTypesName(tumorTypes: TumorType[]) {
     : '';
 }
 
+export function getTumorTypeName(tumorType: TumorType) {
+  return tumorType ? getCancerTypeNameFromOncoTreeType(tumorType) : '';
+}
```
zhx828, line 162: Should we just add the check in getCancerTypeNameFromOncoTreeType?

src/main/webapp/app/shared/api/generated/OncoKbPrivateAPI.ts
```diff
@@ -1051,6 +1053,7 @@ export default class OncoKbPrivateAPI {
         }): Promise < Array < Treatment >
         > {
             return this.searchTreatmentsGetUsingGETWithHttpInfo(parameters).then(function(response: request.Response) {
+                console.log("this is the response: " +  response.body);
```
zhx828, line 1056: A reminder, please remove console logs before merging.

zhx828 (changes requested):
Looks very nice. When no result returned from backend, we need to update the language to say "No cancer type is found"

src/main/webapp/app/pages/HomePage.tsx
Eric-Jiang1: that's for a list of cancer that a drug search returns, do you think we should just keep one list of cancer and determine based on if it came from drug or cancer_type search?

src/main/resources/config/application-dev.yml
```diff
@@ -37,7 +37,8 @@ spring:
       indent-output: true
   datasource:
     type: com.zaxxer.hikari.HikariDataSource
-    url: jdbc:mysql://localhost:3306/oncokb_public?useUnicode=true&characterEncoding=utf8&useSSL=false&useLegacyDatetimeCode=false&serverTimezone=UTC&createDatabaseIfNotExist=true
```
zhx828, line 40: Is there a reason why we want to change the default database name?

src/main/resources/config/application-dev.yml
```diff
@@ -134,7 +135,7 @@ sentry:
 application:
   name: local
   profile: dev
-  api-proxy-url: 'http://localhost:8080/web_war'
```
zhx828, line 137: I think this is updated in the base branch. Could you rebase to the latest?

src/main/resources/config/application-dev.yml
```diff
@@ -149,7 +150,8 @@ application:
     user-registration-channel-id:
   redis:
     # Add the following if you want to opt in to using Redis
-    # enabled: true
+    # Not using Redis
+    enabled: false
```
zhx828, line 154: I don't mind make it more specific. @calvinlu3 any thoughts why we didn't in the first place?

src/main/resources/config/application-dev.yml
Eric-Jiang1: nope, this was when I couldn't set up on my laptop so we tried to see if it was a database connection issue

PR #995, opened 2023-07-05
Support genomic change annotation

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3464 and is related to https://github.com/oncokb/oncokb/pull/3488


src/main/webapp/app/pages/genomicPage/GenomicPage.tsx
```diff
@@ -52,13 +55,32 @@ export default class HgvsgPage extends React.Component<HgvsgPageProps> {
   constructor(props: any) {
     super(props);
 
+    if (props.routing.location.pathname.startsWith(PAGE_ROUTE.HGVSG)) {
```
calvinlu3, line 58: Should we can make the checks case insensitive, otherwise `/HGVSG/:query` would not be annotated.

The other place where I see this: https://www.oncokb.org/YEAR-END-SUMMARY. The tab is not selected.

calvinlu3 (approved):

src/main/webapp/app/pages/genomicPage/GenomicPage.tsx
zhx828: Good point, but I think all of our pages URL are case sensitive. I can fix the `startsWith` usage, we could create a separate issue to discuss whether making URL case insensitive makes sense.

calvinlu3 (approved):
LGTM

PR #1006, opened 2023-07-27
Generate alteration names based on wether to show diff from alteration and name

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3507

### DIFF
--- a/src/main/webapp/app/store/AnnotationStore.ts
+++ b/src/main/webapp/app/store/AnnotationStore.ts
@@ -561,8 +561,8 @@ export class AnnotationStore {
     }
   }
 
-  // need to pass all observables from @computed, so they can bo monitored by mobx
-  computeAlterationName(annotationType, alteration, alterationQuery, annotationData, showDiff: boolean) {
+  // need to pass all observables from @computed, so they can be monitored by mobx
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/store/AnnotationStore.ts
```diff
@@ -562,23 +561,35 @@ export class AnnotationStore {
     }
   }
 
-  @computed
-  get alterationName() {
-    if (this.annotationType === AnnotationType.PROTEIN_CHANGE) {
+  // need to pass all observables from @computed, so they can bo monitored by mobx
```
calvinlu3, line 564:
```suggestion
  // need to pass all observables from @computed, so they can be monitored by mobx
```

src/main/webapp/app/store/AnnotationStore.ts
```diff
@@ -562,23 +561,35 @@ export class AnnotationStore {
...
+  computeAlterationName(annotationType, alteration, alterationQuery, annotationData, showDiff: boolean) {
```
calvinlu3, line 565: Some typescript errors related to the method arguments

calvinlu3 (approved):

PR #1011, opened 2023-08-08
Updates to actionable genes page

zhx828:
- Only show treatment with FDA level. If the treatment is Dx/Px, then the fda level is not available
- Add key to actionable genes page. Multiple tables will be rendered if the key is different. This helps to differentiate the paginated vs non-paginated tables. The react-table lib has issue recognizing remove props.

This fixes https://github.com/oncokb/oncokb/issues/3521


benzuzu (approved):
I haven't actually set up the core before, so I don't know how to test the tables locally (might be good to learn how to do this sometime). The logic makes sense for the fda level though. For multiple tables, when `getTable()` returns the different type of table, does `render()` just remember the old one and render both? And will `render()` be called again when `fdaSectionIsOpen` is changed in the first place?

PR #1012, opened 2023-08-15
Token Format Transition

benzuzu:
See [proposal](https://docs.google.com/document/d/1yLRE2JtKI2T2u5WzO5KS-XzC9YTBplvDeFe40BiTAC0/edit) and [outline](https://docs.google.com/document/d/1Jm2bz27a0pKWk79E72JIi_ndxYjvm1LedulSIQMy_do/edit).

PR #1016, opened 2023-08-28
Add checksum and TokenType into Token class

zhx828:
This is related to https://github.com/oncokb/oncokb-public/pull/1012
The two fields are added through command `jhipster entity Token`. https://www.jhipster.tech/ has been used to generate oncokb-public backend and for both frontend/backend of oncokb-transcript(oncokb-curation). The command will provide all the options for updating the model if exists, or create.


.jhipster/Token.json
```diff
@@ -36,6 +36,20 @@
...
       "fieldType": "Boolean",
       "fieldValidateRules": ["required"]
+    },
+    {
+      "fieldName": "checksum",
+      "fieldType": "String",
+      "fieldValidateRules": ["required", "minlength", "maxlength", "pattern"],
+      "fieldValidateRulesMinlength": "8",
+      "fieldValidateRulesMaxlength": "8",
+      "fieldValidateRulesPattern": "^[a-zA-Z0-9]*$"
```
zhx828, line 46: @benzuzu once merge, we need to update the regex patterns.

zhx828:
The test is currently failing because we need to define the default way to generate checksum and token type when initiating.

PR #1023, opened 2023-09-28
Add hash and search when redirecting from legacy actionable genes page

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3571


src/main/webapp/app/config/constants.tsx
```diff
@@ -534,7 +534,6 @@ export enum PAGE_ROUTE {
   LOGIN = '/login',
   LOGOUT = '/logout',
   LEGACY_DATA_ACCESS = '/dataAccess',
-  DATA_ACCESS = '/data-access',
```
zhx828, line 537: We no longer have data access page. Previously we wanted to redirect to from /dataAccess to /api-access. This was added by mistake.

calvinlu3 (approved):

PR #1025, opened 2023-09-29
Finalize precision oncology therapies page

zhx828:
- Use compact version of the filters(Option 1)
- move drug classification column as first column
- use filtered px tx when calculating drug classification and detection method
- use small button for clear filter and download data


calvinlu3 (approved):
LGTM

PR #1029, opened 2023-10-13
Add rejection email and slack dropdown option due to US sanction

zhx828:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/183


src/main/java/org/mskcc/cbio/oncokb/web/rest/slack/DropdownEmailOption.java
```diff
@@ -292,6 +303,11 @@ public DropdownEmailOptionBuilder isNotModalEmail() {
             return this;
         }
 
+        /**
+         *
+         * @param modalTitle Title showing in the slack modal. It cannot be longer than 15 characters.
```
calvinlu3, line 308: Is the 15 characters a hard requirement? Most of our titles are over 15 character.

calvinlu3 (approved):
Looks good, just a quick question below.

src/main/java/org/mskcc/cbio/oncokb/web/rest/slack/DropdownEmailOption.java
zhx828: Sorry, it should be 25 characters. And it's a hard requirement.

PR #1038, opened 2023-10-23
Rename sanction and license options txt email templates to include postfix

zhx828:
The txt file is used for slack. It seems to be the standard with the postfix(String) added. Need to find out why with Ben


src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
```diff
@@ -94,7 +94,7 @@ public void sendEmail(String to, String from, String cc, String subject, String
         MimeMessage mimeMessage = javaMailSender.createMimeMessage();
         MimeMessageHelper message = new MimeMessageHelper(mimeMessage, isMultipart, StandardCharsets.UTF_8.name());
         message.setTo(to);
-        if (cc != null) {
+        if (StringUtils.isNotEmpty(cc)) {
```
calvinlu3, line 97: Similar to this change, should we update to use `isNotEmpty(sendBy)` on this [line](https://github.com/oncokb/oncokb-public/blob/3695c00c689ed983b2b1404a15669debd24c6631/src/main/java/org/mskcc/cbio/oncokb/service/MailService.java#L193) too?

calvinlu3 (approved):
LGTM

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
zhx828: @calvinlu3 good call. Made the change.

PR #1039, opened 2023-10-24
Show the highest FDA level among evidence

zhx828:
The sortedBy always returns the ascending order and LEVEL_PRIORITY store the lower priority first.

This fixes https://github.com/oncokb/oncokb/issues/3583


calvinlu3 (approved):
Looks good.

PR #1044, opened 2023-11-02
Fix check duplicate users

bprize15:
Fix https://github.com/oncokb/oncokb-pipeline/issues/191

### DIFF
--- a/src/test/java/org/mskcc/cbio/oncokb/service/UserServiceIT.java
+++ b/src/test/java/org/mskcc/cbio/oncokb/service/UserServiceIT.java
@@ -12,7 +12,6 @@ import org.mskcc.cbio.oncokb.service.dto.useradditionalinfo.TrialAccount;
 import org.mskcc.cbio.oncokb.service.mapper.UserMapper;
 
 import io.github.jhipster.security.RandomUtil;
-
 import org.apache.commons.lang3.RandomStringUtils;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mskcc/cbio/oncokb/web/rest/vm/ManagedUserVM.java
```diff
@@ -23,6 +23,8 @@ public class ManagedUserVM extends UserDTO {
 
     private Boolean needsMskRocReview = true;
 
+    private Boolean requestApiAccess;
```
zhx828, line 26: These should not be here.

zhx828 (changes requested):
@bprize15 I mixed the changes related to API change. Could you update?

src/main/java/org/mskcc/cbio/oncokb/web/rest/vm/ManagedUserVM.java
bprize15: Sorry thought I stashed that

src/test/java/org/mskcc/cbio/oncokb/service/UserServiceIT.java
```diff
@@ -12,6 +12,7 @@
 import org.mskcc.cbio.oncokb.service.mapper.UserMapper;
 
 import io.github.jhipster.security.RandomUtil;
+import liquibase.pro.packaged.iu;
```
zhx828, line 15: We don't use this.

zhx828 (approved):
Thanks

PR #1045, opened 2023-11-02
Print error message and slack response if the response code is not 200

zhx828:
Slacked calls failed earlier but we didn't print out the error messages. It's impossible to debug without that.

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
@@ -237,8 +237,11 @@ public class SlackService {
         try {
             WebhookResponse response = slack.send(url, payload);
             log.info("Send the latest user blocks to slack with response code " + response.getCode());
+            if (response.getCode() > 200) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -237,8 +237,11 @@ private void sendBlocks(String url, List<LayoutBlock> layoutBlocks) {
         try {
             WebhookResponse response = slack.send(url, payload);
             log.info("Send the latest user blocks to slack with response code " + response.getCode());
+            if (response.getCode() > 200) {
```
bprize15, line 240: Is != better? It wouldn't really make sense to send any response code <200, but figured != might be better. If not, feel free to disregard.

zhx828: Yeah, I think it makes sense.

bprize15 (approved):
LGTM

PR #1046, opened 2023-11-03
Add role api

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/188
resolves https://github.com/oncokb/oncokb-pipeline/issues/189
resolves https://github.com/oncokb/oncokb-pipeline/issues/193
resolves https://github.com/oncokb/oncokb-pipeline/issues/162


src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -106,6 +106,28 @@ public void sendUserRegistrationToChannel(UserDTO user, boolean trialAccountActi
...
 
+    @Async
+    public void sendUserApiAccessRequestToChannel(UserDTO user) {
+        log.debug("Sending notification to admin group that a user has requested API access");
+        if (StringUtils.isEmpty(this.applicationProperties.getSlack().getUserRegistrationWebhook())) {
+            log.debug("\tSkipped, the webhook is not configured");
+        } else {
+            List<LayoutBlock> layoutBlocks = new ArrayList<>();
+
+            String userPageLink = "(<https://www.oncokb.org/users/" + user.getEmail() + "/|" + user.getEmail() + ">)";
```
zhx828, line 117: We should update to use the base URL. The https://github.com/oncokb/oncokb-public/pull/1047 is merged.

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -274,6 +276,43 @@ public void changePassword(@RequestBody PasswordChangeDTO passwordChangeDto) {
...
+    public void requestApiAccess(@RequestBody ApiAccessRequest apiAccessRequest) {
+        Optional<String> userLogin = SecurityUtils.getCurrentUserLogin();
+        if (userLogin.isPresent()) {
+            Optional<User> user = userService.getUserWithAuthoritiesByLogin(userLogin.get());
+            if (user.isPresent()) {
+                Optional<UserDetailsDTO> userDetailsDTO = userDetailsService.findOneByUser(user.get());
+                if (userDetailsDTO.isPresent()) {
+                    UserDetailsDTO oldUserDetailsDTO = userDetailsDTO.get();
+
+                    AdditionalInfoDTO newAdditionalInfoDTO = oldUserDetailsDTO.getAdditionalInfo();
```
zhx828, line 294: Why don't we modify oldUserDetailsDTO?

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -274,6 +276,43 @@ public void changePassword(@RequestBody PasswordChangeDTO passwordChangeDto) {
...
+                    updatedUserDetailsDTO.setCompanyName(oldUserDetailsDTO.getCompanyName());
+                    updatedUserDetailsDTO.setCity(oldUserDetailsDTO.getCity());
+                    updatedUserDetailsDTO.setCountry(oldUserDetailsDTO.getCountry());
+                    updatedUserDetailsDTO.setAddress(oldUserDetailsDTO.getAddress());
+                    updatedUserDetailsDTO.setAdditionalInfo(newAdditionalInfoDTO);
+                    updatedUserDetailsDTO.setUserId(oldUserDetailsDTO.getUserId());
+                    updatedUserDetailsDTO.setCompanyId(oldUserDetailsDTO.getCompanyId());
+                    
+                    userDetailsService.save(updatedUserDetailsDTO);
+                    slackService.sendUserApiAccessRequestToChannel(userMapper.userToUserDTO(user.get()));//ask about this, probably slower, but sending to slack so no big deal?
```
zhx828, line 310: What does this comment here mean

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
```diff
@@ -92,12 +94,31 @@ public ResponseEntity<String> approveUser(@RequestParam("payload") String action
...
                                 mailService.sendApprovalEmail(userDTO);
                             }
                         }
                         break;
+                    case APPROVE_USER_FOR_API_ACCESS:
+                        boolean apiAccessRequested = userDTO.getAdditionalInfo() != null && userDTO.getAdditionalInfo().getApiAccessRequest() != null && userDTO.getAdditionalInfo().getApiAccessRequest().isRequested();
+                        if (userDTO.getLicenseType() != LicenseType.ACADEMIC || apiAccessRequested) {
+                            Set<String> userDTOAuthorities = userDTO.getAuthorities();
+                            userDTOAuthorities.add(AuthoritiesConstants.API);
+                        }
```
zhx828, line 115: We don't need to check the request anymore here, do we?

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
```diff
@@ -92,12 +94,31 @@ public ResponseEntity<String> approveUser(@RequestParam("payload") String action
...
+
+                        Optional<UserDTO> updatedUser = userService.updateUser(userDTO);
+                        if (updatedUser.isPresent() && updatedUser.get().isActivated()) {
+                            mailService.sendApiAccessApprovalEmail(userDTO);
+                        }
```
zhx828, line 120: Here though, we should make sure only send email if the user does not have API ROLE already, so we don't send the same email multiple times.

src/main/webapp/app/pages/AccountPage.tsx
```diff
@@ -52,11 +57,18 @@ export const InfoRow: React.FunctionComponent<{
...
+  IRegisterProps,
+  IApiAccessRequestState
+> {
   @observable copiedIdToken = false;
+  @observable showConfirmModal = false;
 
   constructor(props: Readonly<IRegisterProps>) {
     super(props);
+    this.state = {
+      apiAccessJustification: '',
```
zhx828, line 70: 1. `apiAccessJustification` does not need to be a state. 
2. We use Mobx observable for state management in Class. See `showConfirmModal`.

src/main/webapp/app/pages/AccountPage.tsx
```diff
@@ -212,37 +224,82 @@ export class AccountPage extends React.Component<IRegisterProps> {
...
-                </div>
-              }
-            >
-              <TokenInputGroups
-                changeTokenExpirationDate={false}
-                tokens={this.tokens}
-                onDeleteToken={this.deleteToken}
-              />
-            </InfoRow>
+            {this.account.authorities.includes(AUTHORITIES.API) ? (
```
zhx828, line 227: We should also handle the case when user has requested the API access. A language like  "We are reviewing your API access request. If you haven't heard back from us after 5 business days, please feel free to contact us." should work.

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
bprize15: That's the my frontend side getting to me, usually try to make all functions pure and recreate rather than modify. But seems there is no reason here, so I will change it

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
bprize15: I needed to map the user to a userDTO, which requires a database query, instead of just passing a user. It seems like the slack service mostly relies on the userDTO object. I could change it to just pass the user, but it would not be consistent with the rest of slack service. I figured since it was just sending a slack message, that extra database read might not matter.

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
bprize15: Yeah there should always be an API access request and the boolean isRequested should always be set. Can additional info ever be null?

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
bprize15: This should only be triggered when the approve button is clicked in slack. The message is only sent to slack if the user has requested API access. The user can only request API access if they do not have ROLE API. I think we can do the check, but is it necessary?

zhx828: I do think it's necessary. The scenario of using the approve button may change, adding an additional check makes sense here.

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
zhx828: Nvm, it's a good point. Referring to my the other comment, I think it's better keep the checking here. However it's the exact same logic of line 98-102. Can we consolidate and create a method?

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
zhx828: I think the current implementation doe snot prevent sending the email multiple times? We should record whether `userDTO` has ROLE_API before `updateUserWithRoleApiIfRequested`. Then use that together with the current if conditions. @bprize15

src/main/java/org/mskcc/cbio/oncokb/service/SlackService.java
```diff
@@ -508,8 +536,14 @@ private List<LayoutBlock> buildUserInfoBlocks(UserDTO user) {
                     userInfo.add(MarkdownTextObject.builder().text("Business Contact Phone:\n" + additionalInfoDTO.getUserCompany().getBusinessContact().getPhone()).build());
                 }
             }
+
+            String apiAccessJustification = "";
+            if (additionalInfoDTO.getApiAccessRequest() != null && additionalInfoDTO.getApiAccessRequest().isRequested()) {
```
calvinlu3, line 541: Could just simplify to below?
```
if (apiAccessRequested) {...}
```

src/main/java/org/mskcc/cbio/oncokb/service/dto/useradditionalinfo/ApiAccessRequest.java
```diff
@@ -0,0 +1,24 @@
+package org.mskcc.cbio.oncokb.service.dto.useradditionalinfo;
+
+public class ApiAccessRequest {
```
calvinlu3, line 3: Implement `Serializable` here?

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
```diff
@@ -53,6 +56,19 @@ public SlackController(UserService userService, UserRepository userRepository, M
         this.userMapper = userMapper;
     }
 
+    private void updateUserWithRoleApiIfRequested(UserDTO userDTO) {
+        boolean apiAccessRequested = userDTO.getAdditionalInfo() != null && userDTO.getAdditionalInfo().getApiAccessRequest() != null && userDTO.getAdditionalInfo().getApiAccessRequest().isRequested();
+        if (userDTO.getLicenseType() != LicenseType.ACADEMIC || apiAccessRequested) {
+            Set<String> userDTOAuthorities = userDTO.getAuthorities();
+            userDTOAuthorities.add(AuthoritiesConstants.API);
+
+            if (userDTO.getLicenseType() == LicenseType.ACADEMIC) {
```
calvinlu3, line 65: Let use `equals()` method when comparing the license types. The `==` checks for reference equality.

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
```diff
@@ -53,6 +56,19 @@ public SlackController(UserService userService, UserRepository userRepository, M
...
+                ApiAccessRequest userApiAccessRequest = userDTO.getAdditionalInfo().getApiAccessRequest();
+                userApiAccessRequest.setRequested(false);
```
calvinlu3, line 67: Minor comment, but probably don't need to create an extra variable here

`userDTO.getAdditionalInfo().getApiAccessRequest().setRequested(false)`

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -274,6 +276,29 @@ public void changePassword(@RequestBody PasswordChangeDTO passwordChangeDto) {
...
     }
 
+    /**
+     * {@code POST  /account/request-api-access} : requests ROLE_API for current user.
+     *
+     * @param apiAccessRequest isRequested boolean and justification (reason for request).
+     */
+    @PostMapping(path = "/account/request-api-access")
+    public void requestApiAccess(@RequestBody ApiAccessRequest apiAccessRequest) {
+        Optional<String> userLogin = SecurityUtils.getCurrentUserLogin();
```
calvinlu3, line 286: We should probably add a check to see if the justification is not empty here using an if condition. You can also add `@NotEmpty` annotation to the ApiAcessRequest's `justification` property.

src/main/webapp/app/components/newAccountForm/NewAccountForm.tsx
```diff
@@ -687,6 +696,68 @@ export class NewAccountForm extends React.Component<INewAccountForm> {
...
               </Row>
             )}
+            {!this.isCommercialLicense && (
+              <Row className={getSectionClassName()}>
+                <Col md="3">
+                  <h5>API Access</h5>
+                </Col>
+                <Col md="9">
+                  <p>
+                    Would you like programmatic access to the OncoKB database
```
calvinlu3, line 706: Replace oncokb text with the `ONCOKB_TM` variable from constants

src/main/webapp/app/pages/AccountPage.tsx
```diff
@@ -134,6 +139,54 @@ export class AccountPage extends React.Component<IRegisterProps> {
...
+        >
+          HERE
+        </span>{' '}
+        to request API access.
+      </span>
+    );
+    const awaitingApiAccess = (
+      <span>
+        We are reviewing your API access request. If you haven't heard back from
+        us after 5 business days, please feel free to contact us.
```
calvinlu3, line 186: Wondering why we say 5 business days to approve api access? Currently for account approval, we say 1-2 business days.

src/main/webapp/app/shared/modal/SimpleConfirmModal.tsx
```diff
@@ -26,6 +27,7 @@ export const SimpleConfirmModal: React.FunctionComponent<{
         </Button>
         <Button
           variant="primary"
+          disabled={props.confirmDisabled}
```
calvinlu3, line 30: If `props.confirmDisabled` is not passed, then we should default to false here

calvinlu3 (commented):
@bprize15 Left some comments and questions. Thanks!

src/main/webapp/app/shared/modal/SimpleConfirmModal.tsx
bprize15: Reactstraps button takes an optional disabled prop as well, so we can just pass it down. If it is null for our component, it will also be null for the button component which is the correct behavior.

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
bprize15: Set @NotEmpty to the justification property so it should never be empty. I think that's enough given an API request cannot come without a justification?

zhx828: @bprize15 does @NonEmpty through exception when it's actually empty? I remember the annotation here mostly for displaying on swagger. It actually doesn't enforce the content to be nonempty.

src/main/webapp/app/shared/modal/SimpleConfirmModal.tsx
zhx828: As logn as @bprize15 you verified the behavior.

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
bprize15: @zhx828 I added `@Valid` to the definition of requestApiAccess method like so:

`@PostMapping(path = "/account/request-api-access") public void requestApiAccess(@Valid @RequestBody ApiAccessRequest apiAccessRequest)`

If the justification is empty, the controller will return 400

calvinlu3: I was thinking that the user could technically use something like postman to send a request with justification set to empty string.

Would like to know if the annotation throws error too.

bprize15: I tested and if they use Postman they will get a 400 response. I also allowed users to send empty justifications temporarily through the website, and the post request was throwing an error

zhx828 (approved):
Thanks!

PR #1047, opened 2023-11-08
Allow configuring app base URL

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3592

### DIFF
--- a/src/main/resources/config/application-dev.yml
+++ b/src/main/resources/config/application-dev.yml
@@ -135,6 +135,7 @@ sentry:
 application:
   name: local
   profile: dev
+  base-url:
   # This is the proxy link to oncokb-core
   api-proxy-url: 'http://localhost:8080/oncokb'
   sitemap-enabled: false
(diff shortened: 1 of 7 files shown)


src/main/resources/config/application-dev.yml
```diff
@@ -135,6 +135,7 @@ sentry:
 application:
   name: local
   profile: dev
+  base-url:
```
calvinlu3, line 138: Should we default to `http://localhost:9095`?

src/main/resources/config/application-prod.yml
```diff
@@ -145,6 +145,7 @@ sentry:
 application:
   name: oncokb
   profile: prod
+  base-url: 'https://www.oncokb.org'
```
calvinlu3, line 148: When we roll our this change, we need to update the config of MSK and EU instances to correct url

calvinlu3 (approved):

bprize15 (approved):

PR #1054, opened 2023-11-15
Add service disruption message

zhx828:
@calvinlu3 let me know whether the time works for you. Also I don't want to show the banner during the AMP conference. Currently it's set up to be displayed on the 19th 0am(GMT) and ends on 27th 0am (GMT).


calvinlu3 (approved):
LGTM. The time is also fine with me

PR #1059, opened 2023-11-28
Do not update API requested after approval && Add title for API request justification in slack

bprize15:
Quick fix allowing ROLE_API to be removed while maintaining requested status in DB


zhx828 (approved):
LGTM, thanks!

PR #1067, opened 2023-12-20
Include all FDA approved oncology therapies

zhx828:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/202


src/main/webapp/app/pages/oncologyTherapiesPage/oncologyTherapiesPage.tsx
```diff
@@ -0,0 +1,679 @@
+import React, { useEffect, useMemo, useState } from 'react';
```
zhx828, line 1: This file is based on deleted precisionOncologyTx page. But there are too many changes, you have to review all unfortunately. @calvinlu3

src/main/webapp/app/components/downloadButton/DownloadButton.tsx
```diff
@@ -17,7 +18,9 @@ export const DownloadButton: React.FunctionComponent<IDownloadButton> = props =>
     <a
       href={href}
       download={download || true}
-      className={`btn btn-primary ${buttonSizeClassName} ${className} ${styles.aTag}`}
+      className={`btn btn-${
+        props.outline ? 'outline-' : ''
+      }primary ${buttonSizeClassName} ${className} ${styles.aTag}`}
```
calvinlu3, line 23: Use `classNames()`. If any of the variables are undefined, then it will add `undefined` to class name.

src/main/webapp/app/index.module.scss
```diff
@@ -22,6 +22,14 @@
   font-size: $bigger-default-font-size;
 }
 
+:global(.btn).actionButton {
```
calvinlu3, line 25: I think we've been using `-` in class names

src/main/webapp/app/pages/oncologyTherapiesPage/oncologyTherapiesPage.tsx
```diff
@@ -0,0 +1,679 @@
...
+        }
+      },
+    },
+    {
+      show: poOnly,
+      Header: (
+        <span>
+          FDA drug label listed biomarker(s)
+          <DefinitionTooltip footnoteKey={'biomarker'} />
+        </span>
```
calvinlu3, line 281: Should sorting in this column be disabled? Doesn't work very well right now since we also have bulleted items in the cell

src/main/webapp/app/index.module.scss
zhx828: This is module class name, in this case, we mostly use camelCase.

PR #1068, opened 2023-12-21
create activation key before sending email

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/219


zhx828 (approved):
Thanks!

PR #1072, opened 2023-12-28
Allow AWS services to get temporary credentials from service account

calvinlu3:
This is related to https://github.com/oncokb/oncokb-pipeline/issues/210

With the new v2 account, we no longer have static AWS access/secret that we can load into our application properties. Instead, we need to run saml2aws to get temporary credentials.

### Approach:
This is based on how the `saml2aws` tool works, which is based on https://aws.amazon.com/blogs/security/how-to-implement-a-general-solution-for-federated-apicli-access-using-saml-2-0/
1. Make a POST request with form data that includes the service account credentials to MSK Idp (https://ssofed.mskcc.org/idp/startSSO.ping?PartnerSpId=urn%3Aamazon%3Awebservices)
2. The response will look something like this:
```
<html>
    <head>
        <title>Submit Form</title>
        <meta name="referrer" content="origin"/>
        <meta http-equiv="x-ua-compatible" content="IE=edge" />
        <meta http-equiv="Content-Security-Policy" content="default-src 'self' 'nonce-ndTSwk4cQBJjyUzM';" />
    </head>
    <body>
       <noscript>
            <p>
                <strong>Note:</strong> Since your browser does not support JavaScript, you must press the Resume button once to proceed.
            </p>
        </noscript>
        <form method="post" action="https://signin.aws.amazon.com/saml">
                        <input type="hidden" name="SAMLResponse" value="[saml response here]"/>
                        <noscript><input type="submit" value="Resume"/></noscript>
        </form>
    </body>
    <script type="text/javascript" nonce="ndTSwk4cQBJjyUzM">
        window.onload = function() {
            document.forms[0].submit()
        }
    </script>

</html>
```
3. Use jsoup (similar to BeautifulSoup in Python) to extract the value of the `input` named `SAMLResponse`
4. Exchange the saml reponse token with AWS Security Token Service to obtain temporary credentials. We are using `StsAssumeRoleWithSamlCredentialsProvider`, which periodically sends an saml assertion to the AWS Security Token Service to maintain short-lived sessions to use for authentication.


src/main/java/org/mskcc/cbio/oncokb/service/SamlService.java
```diff
@@ -0,0 +1,99 @@
...
+            Supplier<AssumeRoleWithSamlRequest> supplier = () ->
+                samlRequest.toBuilder().samlAssertion(getSamlResponse()).build();
+
+            StsAssumeRoleWithSamlCredentialsProvider stsProvider = StsAssumeRoleWithSamlCredentialsProvider
+                .builder()
+                .stsClient(StsClient.builder().credentialsProvider(AnonymousCredentialsProvider.create()).region(Region.US_EAST_1).build())
+                .refreshRequest(supplier)
+                .build();
+            credentialsProvider = stsProvider;
+        }
```
zhx828, line 57: Let's log warning indicating SamlAws is not enabled.

src/main/java/org/mskcc/cbio/oncokb/service/SamlService.java
```diff
@@ -0,0 +1,99 @@
...
+    private void initSecurityTokenServiceProvider() {
+        if (applicationProperties.getSamlAws() != null) {
+            AssumeRoleWithSamlRequest samlRequest = AssumeRoleWithSamlRequest
+                .builder()
+                .principalArn(applicationProperties.getSamlAws().getPrincipalArn())
+                .roleArn(applicationProperties.getSamlAws().getRoleArn())
+                .durationSeconds(28800)
+                .build();
+
+            Supplier<AssumeRoleWithSamlRequest> supplier = () ->
```
zhx828, line 48: Will it throw exception when getSamlResponse returns null?

src/main/java/org/mskcc/cbio/oncokb/service/SamlService.java
```diff
@@ -0,0 +1,99 @@
...
+                samlRequest.toBuilder().samlAssertion(getSamlResponse()).build();
+
+            StsAssumeRoleWithSamlCredentialsProvider stsProvider = StsAssumeRoleWithSamlCredentialsProvider
+                .builder()
+                .stsClient(StsClient.builder().credentialsProvider(AnonymousCredentialsProvider.create()).region(Region.US_EAST_1).build())
```
zhx828, line 53: Should weu use the region from the SamlAwsProp?

zhx828 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/service/SamlService.java
calvinlu3: An exception will be thrown by the aws security token service saying that the saml assertion cannot be null. For `S3Service.getObject()`, the exception will be caught and we currently return an empty optional.

src/main/java/org/mskcc/cbio/oncokb/service/SamlService.java
zhx828: I don't think this is fixed.

zhx828 (approved):
One comment. Once fixed, this is good to go.

PR #1077, opened 2024-01-18
add role api to trial accounts

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3650

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
@@ -376,6 +376,47 @@ public class UserService {
         return user;
     }
 
+    private Optional<UserDTO> updateUserFromUserDTO(UserDTO userDTO) {
+        Optional<UserDTO> updatedUserDTO = Optional.of(userRepository
+        .findById(userDTO.getId()))
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .map(user -> {
+            this.clearUserCaches(user);
+            user.setLogin(userDTO.getLogin().toLowerCase());
+            user.setFirstName(userDTO.getFirstName());
+            user.setLastName(userDTO.getLastName());
+            if (userDTO.getEmail() != null) {
+                user.setEmail(userDTO.getEmail().toLowerCase());
+            }
+            user.setImageUrl(userDTO.getImageUrl());
+            user.setActivationKey(userDTO.getActivationKey());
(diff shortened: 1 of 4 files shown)


src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
```diff
@@ -125,11 +125,21 @@ public ResponseEntity<String> approveUser(@RequestParam("payload") String action
                             mailService.sendApiAccessApprovalEmail(userDTO);
                         }
                         break;
-                    case GIVE_TRIAL_ACCESS:
+                    }
+                    case GIVE_TRIAL_ACCESS: {
```
zhx828, line 129: We don't use braces in switch.

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -438,6 +442,10 @@ public Optional<UserDTO> updateUser(UserDTO userDTO) {
         return updatedUserDTO;
     }
 
+    public Optional<UserDTO> updateUserBeforeTrialAccountActivation(UserDTO userDTO) {
```
zhx828, line 445: What's the purpose of adding this wrapper?

zhx828 (changes requested):
Could you also check why the ROLE_API is not added when creating an account under a company which has a commercial license?

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
bprize15: I added a helper function `updateUserFromUserDTO` which handles the logic of updating the user from the given UserDTO (this was already in `updateUser`). My new function `updateUserBeforeTrialAccountActivation` needs this same logic, but it shouldn't expire the trial account when updating like `updateUser` does. I added this wrapper because I think the name makes the most sense for the use case, and `updateUserFromUserDTO` would not be a descriptive name to expose as an API.

src/main/java/org/mskcc/cbio/oncokb/web/rest/SlackController.java
bprize15: I added these so I could also define `updatedUser` here even though it was defined in the previous statement (since the cases all share the same scope). I can change the name and remove the braces, however.

zhx828 (approved):
Looks good. Thank you.

PR #1081, opened 2024-02-07
Linkout to categorical alteration page if the alteration starts with the categorical alteration

zhx828:
This will trim all additional info after the categorical alteration


calvinlu3 (approved):
Looks good

PR #1083, opened 2024-02-12
Add long text validation and apply it to use case and justification

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3672


bprize15 (approved):
LGTM (besides just updating screenshot)

PR #1090, opened 2024-02-29
Only renew tokens when activating user for PUT endpoint

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3685


zhx828 (approved):
LGTM, thanks!

PR #1092, opened 2024-03-13
ran update oncokb api script

jfkonecn:
oncokb/oncokb#3684


zhx828:
@jfkonecn can you add the changes to https://www.oncokb.org/swagger-ui/index.html#/Cancer%20Genes collection? If you use a random version as request parameter, the response currently doesn't provide any info.

PR #1101, opened 2024-04-03
Add AACR banner

zhx828:


src/main/webapp/app/components/userMessager/UserMessage.tsx
```diff
@@ -118,30 +118,15 @@ if (
...
-          <Linkout
-            link={
-              'https://amp23.eventscribe.net/ajaxcalls/PresentationInfo.asp?PresentationID=1303633'
-            }
-          >
-            Dr. Chakravarty's presentation on OncoKB™
-          </Linkout>{' '}
-          during the plenary session "Clinical Grade Knowledge Bases for Genomic
-          Actionability" (Thu 11/16, 1:15pm MST).
+          Join us at AACR 2024! Visit our poster session (Sat 4/8, 1:30PM,
```
calvinlu3, line 124: Should we add timezone too?

calvinlu3 (approved):

src/main/webapp/app/components/userMessager/UserMessage.tsx
zhx828: Good point!

PR #1106, opened 2024-04-11
company name can be 255 characters on all pages

jfkonecn:
resolves oncokb/oncokb#3727


src/main/webapp/app/components/newAccountForm/NewAccountForm.tsx
```diff
@@ -463,6 +464,9 @@ export class NewAccountForm extends React.Component<INewAccountForm> {
                       ACCOUNT_TITLES.POSITION,
                       this.selectedLicense
                     )}
+                    validate={{
+                      ...OPTIONAL_TEXT_VAL,
+                    }}
```
calvinlu3, line 469: The `jobTitle` field in UserPage.tsx should also have this length validation. Currently it is set to SHORT_TEXT_VAL.

src/main/webapp/app/components/newAccountForm/NewAccountForm.tsx
```diff
@@ -610,7 +614,7 @@ export class NewAccountForm extends React.Component<INewAccountForm> {
                             errorMessage:
                               'Please let us know where you located.',
                           },
-                          ...SHORT_TEXT_VAL,
+                          ...TEXT_VAL,
```
calvinlu3, line 617: The country and city fields in UserPage.tsx also need to be updated to TEXT_VAL

calvinlu3 (changes requested):

src/main/webapp/app/components/newAccountForm/NewAccountForm.tsx
jfkonecn: good catch... I'll fix that

calvinlu3 (approved):

calvinlu3:
@jfkonecn I changed the bug label to fix instead. We use `Fix` for PRs that address the bugs.

PR #1107, opened 2024-04-12
fixed create company users link style

jfkonecn:
fixes oncokb/oncokb#3728

https://github.com/oncokb/oncokb-public/assets/24961694/54dd0268-d5ca-434e-8f7b-dca14c6a4b14

### DIFF
--- a/src/main/webapp/app/pages/companyPage/CompanyPage.tsx
+++ b/src/main/webapp/app/pages/companyPage/CompanyPage.tsx
@@ -519,11 +519,6 @@ export default class CompanyPage extends React.Component<ICompanyPage> {
                           <Link
                             className="btn btn-outline-primary m-2"
                             to={`/companies/${this.company.id}/create-users`}
-                            onClick={() =>
-                              this.props.routing.history.push(
-                                `/companies/${this.company.id}/create-users`
-                              )
-                            }
                           >
                             Create Company Users
                           </Link>
--- a/src/main/webapp/app/pages/companyPage/CompanyPage.tsx
+++ b/src/main/webapp/app/pages/companyPage/CompanyPage.tsx
@@ -516,13 +516,12 @@ export default class CompanyPage extends React.Component<ICompanyPage> {
                       <Col>
                         <div>Quick Tools</div>
                         <div>
-                          <QuickToolButton>
-                            <Link
-                              to={`/companies/${this.company.id}/create-users`}
-                            >
-                              Create Company Users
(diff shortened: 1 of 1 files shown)


src/main/webapp/app/pages/companyPage/CompanyPage.tsx
```diff
@@ -516,13 +516,12 @@ export default class CompanyPage extends React.Component<ICompanyPage> {
...
                         <div>
-                          <QuickToolButton>
-                            <Link
-                              to={`/companies/${this.company.id}/create-users`}
-                            >
-                              Create Company Users
-                            </Link>
-                          </QuickToolButton>
+                          <Link
+                            className="btn btn-outline-primary"
```
calvinlu3, line 520: This button is missing the `m-2` that is used in QuickToolButton. I think we should probably still use that component. To link to another page, you can use the `routing` prop to change the history. See the UserPage.tsx file for an example.

calvinlu3: I think the missing margin is also why the Screenshot Test is also failing.

jfkonecn: I tried using the history prop, but it's breaking the back button.
You not have to hit it twice to get the browser to go back.
Are you sure we need it?

calvinlu3: Oh that's interesting. Have you tried just using the `QuickButtonTool` with an onClick. Maybe adding the history prop call with the `Link` component is causing some issues?

calvinlu3: If that doesn't work, I think the solution you have now is good too.

jfkonecn: What's it for? Shouldn't the like handle the routing history?

calvinlu3: I imagine the QuickToolButton was created so the styling can be reused for all quick tools.

The Link can be used to handle redirecting. I believe the implementation of that component probably pushes to the history to change the location. We can also update the history by directly calling `history.push()`.

I don't see a problem with either approach:
a. Copy quick tool button styles to Link component
b. Or using history to redirect in QuickToolButton component

jfkonecn:
@zhx828, I want to talk to you about using QuickToolButton for links when you get back.

PR #1110, opened 2024-04-23
Slack reports error and limits justification section

jfkonecn:
fixes oncokb/oncokb-pipeline#327


src/main/webapp/app/components/newAccountForm/NewAccountForm.tsx
```diff
@@ -82,6 +83,8 @@ type CompanySelectOptionType = {
   value: CompanyDTO;
 };
 
+export const SLACK_TEXT_VAL = textValidation(2, 1900);
```
calvinlu3, line 86: 1900 is the slack character limit?

calvinlu3 (approved):
LGTM!

PR #1115, opened 2024-05-09
Fix duplicate block id in slack payload

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/350


bprize15 (approved):
LGTM

PR #1120, opened 2024-05-23
Update to API access request feature

zhx828:
- Do not auto approve when academic user requested API access
- Include justification in slack channel
- Show API Access Justification in User page
- Allow admin to change API access justification in User page

This fixes https://github.com/oncokb/oncokb/issues/3738


src/main/webapp/app/pages/userPage/UserPage.tsx
```diff
@@ -412,13 +412,30 @@ export default class UserPage extends React.Component<IUserPage> {
...
+        this.user.authorities.includes(AUTHORITIES.API) &&
+        !values.authorities.includes(AUTHORITIES.API)
+      ) {
         updatedAdditionalInfo.apiAccessRequest = {
           requested: false,
           justification: '',
         };
       }
 
+      if (values.additionalInfoApiAccessJustification) {
```
bprize15, line 427: So if an admin modifies the justification, and API request will automatically be created? Could this be an issue if we want to hold onto their request justification, but an admin modifies it, generating a new request?

PR #1121, opened 2024-05-28
Annotation visualisation

aishikanandi:
This is the first draft of the visualisation component for OncoKB annotations. 

To do:
- [ ] Tabs for annotation type
- [ ] Publish OncoKB reused components to decrease the package size 
- [ ] Dropdown for columns


src/main/webapp/app/oncokb-annotation-visualisation/AnnotationVisualisation.tsx
```diff
@@ -0,0 +1,380 @@
...
+  isPatientInfoVisible?: boolean;
+  notifications: NotificationImplication[];
+}
+import { annotationColumns, treatmentColumns } from './config/constants';
+
+export interface AnnotationVisualisationState {
+  selectedAnnotationColumns: string[];
+  selectedTreatmentColumns: string[];
+}
+export class AnnotationVisualisation extends React.Component<
```
bprize15, line 39: Could you make this a function component. So just `export function AnnotationVisualisation() { ... }`

src/main/webapp/app/oncokb-annotation-visualisation/AnnotationVisualisation.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+  componentDidMount() {
+    const savedAnnotationColumns = localStorage.getItem(
+      'selectedAnnotationColumns'
+    );
+    const savedTreatmentColumns = localStorage.getItem(
+      'selectedTreatmentColumns'
+    );
+
+    if (savedAnnotationColumns && savedTreatmentColumns) {
```
bprize15, line 122: Can only one of them be set? What happens in that case?

src/main/webapp/app/oncokb-annotation-visualisation/AnnotationVisualisation.tsx
```diff
@@ -0,0 +1,380 @@
...
+  get annotationTableColumns(): SearchColumn<AnnotationImplication>[] {
+    const selectedKeys = this.state.selectedAnnotationColumns;
+    const filteredColumns = annotationColumns.filter(col =>
+      selectedKeys.includes(col.key)
+    );
+
+    return filteredColumns.map(column => ({
+      ...getDefaultColumnDefinition(
+        column.key,
+        1200 / this.state.selectedAnnotationColumns.length
```
bprize15, line 184: I wouldn't just make every column have an even portion of the table width. Also you assume the table width is always 1200 I imagine. What about on changing screen sizes?

src/main/webapp/app/oncokb-annotation-visualisation/components/Utils.tsx
```diff
@@ -0,0 +1,929 @@
...
+    <i className={`oncokb icon ${LevelOfEvidenceToClassnames(level)}`} />
+  );
+  return withDescription ? (
+    <LevelWithDescription level={level}>{oncokbIcon}</LevelWithDescription>
+  ) : (
+    oncokbIcon
+  );
+};
+
+export function getDefaultColumnDefinition<T>(
```
bprize15, line 523: Not using template parameter

src/main/webapp/app/oncokb-annotation-visualisation/AnnotationVisualisation.tsx
```diff
@@ -0,0 +1,380 @@
...
+        option.value ? option.value : option.prop
+      ),
+    });
+    localStorage.setItem(
+      'selectedAnnotationColumns',
+      JSON.stringify(selectedOptions.map((option: any) => option.value))
+    );
+  };
+
+  handleTreatmentColumnsChange = (selectedOptions: any) => {
```
bprize15, line 142: Can you use the type instead of `any`?

bprize15: This comment applies above as well and in most places you see `any` unless there is a good reason to use it

src/main/webapp/app/oncokb-annotation-visualisation/AnnotationVisualisation.tsx
```diff
@@ -0,0 +1,380 @@
...
+    const selectedKeys = this.state.selectedTreatmentColumns;
+    const filteredColumns = treatmentColumns.filter(col =>
+      selectedKeys.includes(col.key)
+    );
+    const hasAnnotation = selectedKeys.some(
+      column => column === TREATMENTS_TABLE_COLUMN_KEY.ANNOTATION
+    );
+    const totalColumns = selectedKeys.length;
+    const calculatedWidth = hasAnnotation
+      ? 800 / (totalColumns - 1)
```
bprize15, line 165: Again, I'd be careful with these hardcoded widths. See my other comment for details.

src/main/webapp/app/oncokb-annotation-visualisation/AnnotationVisualisation.tsx
```diff
@@ -0,0 +1,380 @@
...
+  filterTreatmentsByType(alterationType: string): TreatmentImplication[] {
+    return this.props.treatments.filter(
+      treatment => treatment.alterationType === alterationType
+    );
+  }
+
+  render() {
+    return (
+      <>
+        {this.props.isPatientInfoVisible && (
```
bprize15, line 206: It looks like `patientId` and `patientInfo` are only used when `isPatientInfoVisible === true`. However, the first two props I mentioned are not optional. I would prefer to have these bundled into an optional object like the following:

```
patientInfo?: {
    id: string
    info: string
}
```

`id` and `info` could also be optional.

src/main/webapp/app/oncokb-annotation-visualisation/AnnotationVisualisation.tsx
```diff
@@ -0,0 +1,380 @@
...
+              <h2 className="mb-1" style={{ color: COLOR_BLUE }}>
+                {this.props.patientId}
+              </h2>
+            </div>
+            <div className="flex flex-col">
+              <h6>{this.props.patientInfo}</h6>
+            </div>
+          </div>
+        )}
+        <Tabs
```
bprize15, line 218: Just wondering why you didn't use the `Tabs` component we already had?

PR #1124, opened 2024-06-14
Update reference regex to support 1 level of nested parenthesis

calvinlu3:
This is related to https://github.com/oncokb/oncokb-pipeline/issues/359


zhx828 (approved):
LGTM

PR #1125, opened 2024-06-17
Limit request methods on the proxy endpoint

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3711


zhx828 (approved):
Looks good and tested. Thanks!

PR #1127, opened 2024-06-18
Center align navbar text and remove account text for screen less than 1500px

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3742

### DIFF
--- a/src/main/webapp/app/components/Header.tsx
+++ b/src/main/webapp/app/components/Header.tsx
@@ -234,10 +234,7 @@ class Header extends React.Component<IHeaderProps> {
                     <AccountMenu
                       isAuthenticated={this.props.isUserAuthenticated}
                       isAdmin={this.props.isAdmin}
-                      hideAccountText={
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/components/Header.tsx
```diff
@@ -234,6 +234,10 @@ class Header extends React.Component<IHeaderProps> {
                     <AccountMenu
                       isAuthenticated={this.props.isUserAuthenticated}
                       isAdmin={this.props.isAdmin}
+                      hideAccountText={
+                        this.props.windowStore.isLargeScreen &&
+                        !this.props.windowStore.isXLscreen
```
zhx828, line 239: Should this just be `!this.props.windowStore.isXLscreen`?

calvinlu3: Yeah, I think that is a better condition. I just changed the prop to `showAccountText` and the condition to `isXLscreen`.

zhx828 (approved):
LGTM. thanks

PR #1133, opened 2024-07-09
Added unit test to usage metrics endpoints

jfkonecn:
I created a private `MockS3Data` class with a `files` property which generates and stores the mock JSON S3 files and the expected responses from all the endpoints. I hard coded a seed for the random number generator so that the tests always have the same data passed to them which should make it easier to debug any breaks.

I made the class private since it seemed unnecessary to create file for a class that is only needed once. 

I think I may have caught some bugs with some endpoints, but I opted to maintain the same behavior so the current expected response data may not look quite right. My goal is to make it obvious what the behavior changes are when I make the 3 year timeline change in `UsageAnalysisController`.

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/config/ClockConfig.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/config/ClockConfig.java
@@ -1,0 +1,14 @@
+package org.mskcc.cbio.oncokb.config;
+
+import java.time.Clock;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ClockConfig {
+
+  @Bean
+  public Clock clock() {
+    return Clock.systemDefaultZone();
+  }
+}
--- a/src/main/java/org/mskcc/cbio/oncokb/config/ClockConfig.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/config/ClockConfig.java
@@ -9,6 +9,6 @@ public class ClockConfig {
 
   @Bean
   public Clock clock() {
-    return Clock.systemDefaultZone();
+    return Clock.systemUTC();
   }
 }
--- a/src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
+++ b/src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
@@ -203,13 +203,13 @@ public class UsageAnalysisControllerIT {
 
   private static class MockS3Data {
 
-    private static class Tuple<X, Y> {
-      public final X first;
-      public final Y second;
+    private static class KeyValuePair<Key, Value> {
+      public final Key key;
+      public final Value value;
 
-      public Tuple(X first, Y second) {
-        this.first = first;
-        this.second = second;
+      public KeyValuePair(Key key, Value value) {
+        this.key = key;
+        this.value = value;
       }
     }
 
@@ -242,7 +242,13 @@ public class UsageAnalysisControllerIT {
       return sampleUrls[Math.abs(random.nextInt() % sampleUrls.length)];
     }
 
-    private Tuple<String, JsonObject> fetchJsonObjectFromFilesObject(
+    /**
+     * Get value from files JSON object based on the JSON path created by the
+     * {@code keys}. If an object is missing in the JSON path then it's created.
+     * @param keys The keys in JSON path
+     * @return The resulting {@code KeyValuePair}
+     */
+    private KeyValuePair<String, JsonObject> fetchJsonObjectFromFilesObject(
       String... keys
     ) {
       JsonObject obj = files;
@@ -254,10 +260,19 @@ public class UsageAnalysisControllerIT {
         obj = (JsonObject) obj.get(key);
       }
       String lastKey = keys[keys.length - 1];
-      return new Tuple<>(lastKey, obj);
+      return new KeyValuePair<>(lastKey, obj);
     }
 
-    private Tuple<String, JsonObject> fetchJsonObjectFromFilesObject(
+    /**
+     * Get value from files JSON object based on the JSON path created by the
+     * {@code filePath}/[{@code index}]/{@code keys}.
+     * If an object is missing in the JSON path then it's created.
+     * @param filePath First key in the files object.
+     * @param index Array index of the array the {@code filePath} parameter points to.
+     * @param keys The remaining JSON path keys.
+     * @return The resulting {@code KeyValuePair}
+     */
+    private KeyValuePair<String, JsonObject> fetchJsonObjectFromFilesObject(
       String filePath,
       int index,
       String... keys
@@ -283,13 +298,19 @@ public class UsageAnalysisControllerIT {
       if (!obj.has(lastKey)) {
         obj.addProperty(lastKey, 0);
       }
-      return new Tuple<>(lastKey, obj);
+      return new KeyValuePair<>(lastKey, obj);
     }
 
+    /**
+     * Adds the passed value to the value found in the files JSON object based on the JSON path created by the
+     * keys. If an object is missing in the JSON path then it's created.
+     * @param value Adds the specified value at the given JSON path.
+     * @param keys The keys in JSON path.
+     */
     private void safeAddNestedValueInFilesObject(int value, String... keys) {
-      Tuple<String, JsonObject> tuple = fetchJsonObjectFromFilesObject(keys);
-      String lastKey = tuple.first;
-      JsonObject obj = tuple.second;
+      KeyValuePair<String, JsonObject> pair = fetchJsonObjectFromFilesObject(keys);
+      String lastKey = pair.key;
+      JsonObject obj = pair.value;
       if (!obj.has(lastKey)) {
         obj.addProperty(lastKey, 0);
       }
@@ -298,67 +319,105 @@ public class UsageAnalysisControllerIT {
       obj.addProperty(lastKey, number + value);
     }
 
+    /**
+     * Adds the passed value to the value found in the files JSON object based on the JSON path created by the
+     * keys. If an object is missing in the JSON path then it's created.
+     * @param value Adds the specified value at the given JSON path.
+     * @param filePath First key in the files object.
+     * @param index Array index of the array the {@code filePath} parameter points to.
+     * @param keys The remaining JSON path keys.
+     */
     private void safeAddNestedValueInFilesObject(
       int value,
       String filePath,
       int index,
       String... keys
     ) {
-      Tuple<String, JsonObject> tuple = fetchJsonObjectFromFilesObject(
+      KeyValuePair<String, JsonObject> pair = fetchJsonObjectFromFilesObject(
         filePath,
         index,
         keys
       );
-      String lastKey = tuple.first;
-      JsonObject obj = tuple.second;
+      String lastKey = pair.key;
+      JsonObject obj = pair.value;
       JsonPrimitive primitive = (JsonPrimitive) obj.get(lastKey);
       int number = primitive.getAsInt();
       obj.addProperty(lastKey, number + value);
     }
 
+    /**
+     * Set the passed value to the value found in the files JSON object based on the JSON path created by the
+     * keys. If an object is missing in the JSON path then it's created.
+     * @param value Sets the specified value at the given JSON path.
+     * @param keys The keys in JSON path.
+     */
     private void safeSetNestedValueInFilesObject(String value, String... keys) {
-      Tuple<String, JsonObject> tuple = fetchJsonObjectFromFilesObject(keys);
-      String lastKey = tuple.first;
-      JsonObject obj = tuple.second;
+      KeyValuePair<String, JsonObject> pair = fetchJsonObjectFromFilesObject(keys);
+      String lastKey = pair.key;
+      JsonObject obj = pair.value;
       obj.addProperty(lastKey, value);
     }
 
+    /**
+     * Set the passed value to the value found in the files JSON object based on the JSON path created by the
+     * keys. If an object is missing in the JSON path then it's created.
+     * @param value The value that is going to be added to value located in the JSON path.
+     * @param filePath First key in the files object.
+     * @param index Array index of the array the {@code filePath} parameter points to.
+     * @param keys The remaining JSON path keys.
+     */
     private void safeSetNestedValueInFilesObject(
       String value,
       String filePath,
       int index,
       String... keys
     ) {
-      Tuple<String, JsonObject> tuple = fetchJsonObjectFromFilesObject(
+      KeyValuePair<String, JsonObject> pair = fetchJsonObjectFromFilesObject(
         filePath,
         index,
         keys
       );
-      String lastKey = tuple.first;
-      JsonObject obj = tuple.second;
+      String lastKey = pair.key;
+      JsonObject obj = pair.value;
       obj.addProperty(lastKey, value);
     }
 
+      /**
+       * Set the passed value to the value found in the files JSON object based on the JSON path created by the
+       * keys. If an object is missing in the JSON path then it's created.
+       * @param value The value that is going to be added to value located in the JSON path.
+       * @param filePath First key in the files object.
+       * @param index Array index of the array the {@code filePath} parameter points to.
+       * @param keys The remaining JSON path keys.
+       */
     private void safeSetNestedValueInFilesObject(
       float value,
       String filePath,
       int index,
       String... keys
     ) {
-      Tuple<String, JsonObject> tuple = fetchJsonObjectFromFilesObject(
+      KeyValuePair<String, JsonObject> pair = fetchJsonObjectFromFilesObject(
         filePath,
         index,
         keys
       );
-      String lastKey = tuple.first;
-      JsonObject obj = tuple.second;
+      String lastKey = pair.key;
+      JsonObject obj = pair.value;
       obj.addProperty(lastKey, value);
     }
 
+     /**
+     * Set an empty object to the value found in the files JSON object based on the JSON path created by the
+     * keys. If an object is missing in the JSON path then it's created.
+     * @param value The value that is going to be added to value located in the JSON path.
+     * @param filePath First key in the files object.
+     * @param index Array index of the array the {@code filePath} parameter points to.
+     * @param keys The remaining JSON path keys.
+     */
     private void safeSetNestedEmptyObjectInFilesObject(String... keys) {
-      Tuple<String, JsonObject> tuple = fetchJsonObjectFromFilesObject(keys);
-      String lastKey = tuple.first;
-      JsonObject obj = tuple.second;
+      KeyValuePair<String, JsonObject> pair = fetchJsonObjectFromFilesObject(keys);
+      String lastKey = pair.key;
+      JsonObject obj = pair.value;
       if (!obj.has(lastKey)) {
         obj.add(lastKey, new JsonObject());
       }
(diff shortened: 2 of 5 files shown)


src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,745 @@
...
+        content().contentTypeCompatibleWith(MediaType.APPLICATION_JSON)
+      )
+      .andExpect(content().json(expected, true));
+  }
+
+  private static class MockS3Data {
+
+    private static class Tuple<X, Y> {
+      public final X first;
+      public final Y second;
```
zhx828, line 208: Give some meaningful names here? It's used by this test anyway, may as well just use lastKey and jsonObject?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,745 @@
...
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+import org.testcontainers.shaded.org.apache.commons.io.IOUtils;
+import software.amazon.awssdk.core.ResponseInputStream;
+import software.amazon.awssdk.http.AbortableInputStream;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
+
+/**
+ * Unit tests for the {@link UsageAnalysisController } REST controller.
+ */
+public class UsageAnalysisControllerIT {
```
zhx828, line 48: I have to say this file is very hard to review. Maybe add some comments and wrap some code together for better readability?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,745 @@
...
+    private CompanyDTO createMockCompany(int companyIndex, String companyName) {
+      CompanyDTO companyDTO = new CompanyDTO();
+      companyDTO.setName(companyName);
+      companyDTO.setId((long) companyIndex);
+      companyDTO.setLicenseType(LicenseType.ACADEMIC);
+      return companyDTO;
+    }
+
+    private User createMockUser(
+      int userIndex,
```
zhx828, line 733: Nit picking, should this be userId?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,745 @@
...
+      String userName,
+      String companyName
+    ) {
+      User user = new User();
+      user.setId((long) userIndex);
+      user.setEmail(userName + "@" + companyName + ".com");
+      user.setFirstName(String.valueOf(userName.charAt(userName.length() - 1)));
+      user.setLastName(userName.substring(0, userName.length() - 1));
```
zhx828, line 741: I plan to make a comment about the logic of parsing username to F/L. But I guess it doesn't matter because we don't really care how user is named?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,745 @@
...
+      userDto.setLastName(user.getLastName());
+      userDto.setEmail(user.getEmail());
+      userDto.setJobTitle(user.getLastName() + " Job Title");
+      userDto.setCompany(companyDTO);
+      userDto.setCompanyName(companyDTO.getName());
+      userDto.setLicenseType(LicenseType.ACADEMIC);
+      return userDto;
+    }
+
+    private CompanyDTO createMockCompany(int companyIndex, String companyName) {
```
zhx828, line 724:
```suggestion
    private CompanyDTO createMockCompany(int companyId, String companyName) {
```

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,745 @@
...
+    public MockS3Data(int seed, String[] userNames) throws Exception {
+      random = new Random(seed);
+      addUsers(userNames);
+    }
+
+    private String getNextSampleUrl() {
+      return sampleUrls[Math.abs(random.nextInt() % sampleUrls.length)];
+    }
+
+    private Tuple<String, JsonObject> fetchJsonObjectFromFilesObject(
```
zhx828, line 245: I'm a bit confused about this method. This is to fetch the last key json but backfill all keys if they do not exist in obj?  Maybe add some comments here?

zhx828 (changes requested):
Not completed done with my review but like to hear from you on my comments before moving on further. Thanks!

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: It's use by the logic that creates the mock json file, but I'll change it to `KeyValuePair`

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: It doesn't matter too much, but the way I have the user name setup is last name ending with the first initial. So `doej` break down into `doe` and `j`.

The only reason why it sort of matter is that if the data is too random then it can be challenging to debug a test so I wanted to have realistic names.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: That is correct, but I'll add some comments.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+                if (isThisYear) {
+                  updateResourceSummaryExpectedResponse(
+                    value,
+                    oncokbEndpoint,
+                    monthKey
+                  );
+                }
+
+                // update expected response for user usage endpoint
+                if (isThisYear) {
```
jfkonecn, line 548: @zhx828 I know limiting the usage metrics to only "this" year isn't what we want, but it is currently how the code works.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+      }
+      String lastKey = keys[keys.length - 1];
+      return new KeyValuePair<>(lastKey, obj);
+    }
+
+    /**
+     * Get value from files JSON object based on the JSON path created by the
+     * {@code filePath}/[{@code index}]/{@code keys}.
+     * If an object is missing in the JSON path then it's created.
+     * @param filePath First key in the files object.
```
zhx828, line 273: Is the description for filePath accurate?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+      obj = (JsonObject) arr.get(index);
+
+      for (int i = 0; i < keys.length - 1; i++) {
+        String key = keys[i];
+        if (!obj.has(key)) {
+          obj.add(key, new JsonObject());
+        }
+        obj = (JsonObject) obj.get(key);
+      }
+      String lastKey = keys[keys.length - 1];
```
zhx828, line 300: Do you intend to let the test fails by throwing out of bound exception when no key been added to this method?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+      }
+      JsonArray arr = (JsonArray) obj.get(filePath);
+      while (index >= arr.size()) {
+        arr.add(new JsonObject());
+      }
+      obj = (JsonObject) arr.get(index);
+
+      for (int i = 0; i < keys.length - 1; i++) {
+        String key = keys[i];
+        if (!obj.has(key)) {
```
zhx828, line 295: When key doesn't exist, should we directly return 0 instead of continuing?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+          DateTimeFormatter dayFormatter = DateTimeFormatter.ofPattern(
+            "yyyy-MM-dd"
+          );
+          DateTimeFormatter monthFormatter = DateTimeFormatter.ofPattern(
+            "yyyy-MM"
+          );
+
+          String yearKey = String.valueOf(year);
+
+          for (int month = 1; month <= 12; month++) {
```
zhx828, line 480: Interesting testing stratagy. I'm not testing expert. But would dynamic mock data introduces unpredictability?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
zhx828: You expect to change this after finishing "3 years" requirement to make the test pass?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+        )];
+    }
+
+    /**
+     * Get value from files JSON object based on the JSON path created by the
+     * {@code keys}. If an object is missing in the JSON path then it's created.
+     * @param keys The keys in JSON path
+     * @return The resulting {@code KeyValuePair}
+     */
+    private KeyValuePair<String, JsonObject> fetchJsonObjectFromFilesObject(
```
zhx828, line 254: Two fetchJsonObjectFromFilesObject methods have the same name and same return but produce completely different JSONObject, one with last index included and set the initial value and one without last index. 
Can we maybe update the comment and even the method name?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+     */
+    private void safeAddNestedValueInFilesObject(int value, String... keys) {
+      KeyValuePair<String, JsonObject> pair = fetchJsonObjectFromFilesObject(
+        keys
+      );
+      String lastKey = pair.key;
+      JsonObject obj = pair.value;
+      if (!obj.has(lastKey)) {
+        obj.addProperty(lastKey, 0);
+      }
```
zhx828, line 321: Why isn't this part of `fetchJsonObjectFromFilesObject`? The other `fetchJsonObjectFromFilesObject` method does have it.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+      obj.addProperty(lastKey, number + value);
+    }
+
+    /**
+     * Set the passed value to the value found in the files JSON object based on the JSON path created by the
+     * keys. If an object is missing in the JSON path then it's created.
+     * @param value Sets the specified value at the given JSON path.
+     * @param keys The keys in JSON path.
+     */
+    private void safeSetNestedValueInFilesObject(String value, String... keys) {
```
zhx828, line 359: You added `safe` here. Do you imply a logic which could be not safe?

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+    private void updateSpecificEndpointUserUsageExpectedResponse(
+      int value,
+      User user,
+      String oncokbEndpoint,
+      String monthKey
+    ) {
+      String usageResourcesEndpoint = "/api/usage/resources";
+      safeAddNestedValueInFilesObject(
+        value,
+        usageResourcesEndpoint + "?endpoint=" + oncokbEndpoint,
```
zhx828, line 658: This can be a constant within this method.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,931 @@
...
+  public void setup() throws Exception {
+    usageAnalysisController = new UsageAnalysisController();
+    MockitoAnnotations.initMocks(this);
+    this.mockS3ObjectResponse(data);
+    this.restMockMvc =
+      MockMvcBuilders.standaloneSetup(usageAnalysisController).build();
+  }
+
+  @Test
+  public void shouldGetUsageSummaryResources() throws Exception {
```
zhx828, line 134: Am I correct to say these tests are to test whether the endpoints will return previously generated files properly?

zhx828 (changes requested):
All places where you do `keys[keys.length - 1]`, you should expect out of bound exception. I left a comment asking whether this is your intention. If not, we need to handle it. Out of bound exception is not the best way to show why the test fail I suppose?

And thank you for wrapping logics into methods, so much easier to review!

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: Yes

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: It shouldn't happen since this code is just generating the mock data. I don't think people will touch this code very often, but I'll update the methods so that this exception cannot occur.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: Basically I'm doing the equivalent of a `mkdir -p` here. I want to make sure the entire path exists and also get the value from the path.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: Yes, but I'm hard coding the seed for the random number generator so the data is always the same unless the calls to the random number change which would only happen if we make changes to the mock data.

I thought about making just a bunch of JSON files, but I felt like this would be easier to make changes to.

That said, I could switch this to reading a bunch of JSON files instead of dynamically generating the mock data if you feel it's better to maintain.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: I called it `safe` because if there are missing objects then they are created for you instead of an exception being thrown.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: Yeah they're to test what the current behavior of the endpoints are.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: I'll remove the logic to set the initial value to avoid confusion.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
zhx828: The data is not always the same because you also use `today` as timestamp. The file will be different every day. Is that ok given the use case here? If it's ok, fine to me to keep the current logic.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,957 @@
...
+        String key = restOfTheKeys[i];
+        if (!obj.has(key)) {
+          obj.add(key, new JsonObject());
+        }
+        obj = (JsonObject) obj.get(key);
+      }
+      String lastKey = restOfTheKeys.length == 0
+        ? firstKey
+        : restOfTheKeys[restOfTheKeys.length - 1];
+      return new KeyValuePair<>(lastKey, obj);
```
zhx828, line 312: This is the dup of 263-275

zhx828 (commented):
Left few comments.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: I forgot about that part, but since the endpoint logic is dependent on today's date there isn't a way I can think of that wouldn't require changes to `UsageAnalysisController`. You'd have to mock out all library calls to `java.time`.

I don't think it's worth mocking that logic since this will only shift the data daily.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
```diff
@@ -0,0 +1,957 @@
...
+     * {@code firstKey}/[{@code index}]/{@code keys}.
+     * If an object is missing in the JSON path then it's created.
+     * @param firstKey First key in the files object.
+     * @param index Array index of the array the {@code firstKey} parameter points to.
+     * @param restOfTheKeys The remaining JSON path keys.
+     * @return The resulting {@code KeyValuePair}
+     */
+    private KeyValuePair<String, JsonObject> safeFetchJsonObjectFromFilesObject(
+      String firstKey,
+      int index,
```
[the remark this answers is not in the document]
jfkonecn: Kind of. It's assuming the first key points to an array of JSON objects.

Yes, it is assumed that the rest of the values are JSON objects.

It is risky to assume there will be an array, but it's only being used for this test code. The class is also scoped to just this test class so there isn't a risk someone will use it outside of this class unless they change the code. I don't it's worth making this code more generic if this is only place it's used.

If we want absolute type safety then I should switch to classes instead of building out JSON.

src/test/java/org/mskcc/cbio/oncokb/web/rest/UsageAnalysisControllerIT.java
jfkonecn: I figured it out. I can use the built-in java clock object.

src/main/java/org/mskcc/cbio/oncokb/config/ClockConfig.java
```diff
@@ -0,0 +1,14 @@
...
+import java.time.Clock;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ClockConfig {
+
+  @Bean
+  public Clock clock() {
+    return Clock.systemDefaultZone();
```
zhx828, line 12: Can we make sure this is UTC, if not we should make it be.

PR #1134, opened 2024-07-18
Package Size Optimization POC

jfkonecn:
Reference Links
[Dynamic Imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports)
[Webpack Lazy Loading](https://webpack.js.org/guides/lazy-loading/)
[React Lazy Loading](https://legacy.reactjs.org/docs/code-splitting.html)

### DIFF
--- a/package.json
+++ b/package.json
@@ -168,6 +168,7 @@
     "ts-loader": "7.0.1",
     "typescript": "3.8.3",
     "webpack": "4.43.0",
+    "webpack-bundle-analyzer": "^4.10.2",
     "webpack-cli": "3.3.11",
     "webpack-dev-server": "3.10.3",
     "webpack-merge": "4.2.2",
--- a/src/main/webapp/app/routes/AdminRoutes.tsx
+++ b/src/main/webapp/app/routes/AdminRoutes.tsx
@@ -1,16 +1,28 @@
-import React from 'react';
+import React, { lazy } from 'react';
 import { ErrorBoundaryRoute } from 'app/shared/error/error-boundary-route';
 import { PAGE_ROUTE } from 'app/config/constants';
 import { Switch } from 'react-router-dom';
 import PageNotFound from 'app/shared/error/page-not-found';
-import SendEmailsPage from 'app/pages/sendEmails/SendEmailsPage';
-import UserDetailsPage from '../pages/userManagement/UserDetailsPage';
+const SendEmailsPage = lazy(
+  () => import('app/pages/sendEmails/SendEmailsPage')
+);
+const UserDetailsPage = lazy(
+  () => import('../pages/userManagement/UserDetailsPage')
+);
 import { CreateAccountPage } from 'app/pages/CreateAccountPage';
-import UsageAnalysisPage from 'app/pages/usageAnalysisPage/UsageAnalysisPage';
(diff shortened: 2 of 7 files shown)


zhx828:
TODO list:
- find the list of low hanging fruits, we do a quick fix
- establish all potential big decrease that we can do
- identify the things we can not do

jfkonecn:
@zhx828 

# Optimize Public Bundle

## Context

Most of OncoKB's pages are not being index by Google's SEO. They are getting the following error, "Duplicate without user-selected canonical". A canonical link has been added to OncoKB's head element, but the pages are still not being
indexed. A Google Lighthouse run was done in production and the performance was given a score of 51 out of 100.
It was concluded that the reason why Google is not indexing OncoKB's pages is that the pages are taking too long to load which is confusing Google's crawler and making it think that there are a lot of duplicate pages.

The lighthouse report's recommendation to improve performance is to reduce the size of JavaScript files being download on the client.

After running the [Webpack Bundle Analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) in production mode `vendors.chuck.js` is the largest file at 7.7MB uncompressed. Here is a breakdown of the largest packages:

1. `react-mutation-mapper` 791 KB
2. `cbioportal-frontend-commons` 588 KB
3. `@sentry` 735 KB
4. `lodash.js` 640 KB
5. `core-js` 373 KB
6. `victory-core/es` 325 KB
7. `jquery.js` 301 KB
8. `reactstrap` 271 KB
9. `xregexp` 220 KB
10. `sanitize-html` 194 KB
11. `react-bootstrap` 173 KB
12. `@hot-loader/react-dom` 172 KB
13. `react-select` 152 KB
14. `moment.js` 142 KB
15. `mobx/lib` 135 KB

NOTE: Chunking can be used to split the JavaScript files into multiple
smaller files using [dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports)
or [react lazy loading](https://legacy.reactjs.org/docs/code-splitting.html).

## Options

### Remove Unnecessary Packages

Easiest to implement, but it may be difficult to identify enough packages to remove.

### Lazy Load Large Packages

Slightly harder than removing packages, but will require using promises to get packages which will complicate the code.

### Use React Server Components

This will require an infrastructure change since there will need to be a node server running, but will allow for server side rendering of webpages, so it should reduce the bundle size of JavaScript being downloaded.

### Use [Amazon CloudFront](https://aws.amazon.com/cloudfront/) to Serve the Frontend

Since the frontend will be in a CDN there should be a performance improve regarding latency, but will require an infrastructure change. The package size may need to be reduced anyway.

### Use Spring to Server Side Render Performance Critical Pages

We could use spring to server side render webpages which will make eliminate the need for a massive JavaScript bundle size. React could still be used for pages, such as the admin pages, which do not need to be indexed by google, but this
will be a massive code change.

### Complete Rewrite with Server Side Rendering

Avoid React all together and try to keep the JavaScript as small as possible.

zhx828:
Some thoughts(will be updated continuously):
- Lodash: We should just get rid it. lodash repo is no longer maintained, the package size is huge, most of the methods can be replaced with native methods. 
- Sentry: we can do layzy loading which the error report will still be included in the initial load. https://togithub.com/getsentry/sentry-javascript/issues/2707

PR #1141, opened 2024-08-05
Remove lodash and use native methods instead

zhx828:
Some of the methods can be replaced easily in place, but some have been replaced with helper methods which are based on es6 native methods.

This change will not remove lodash from the dependency. We still have two packages heavily rely on lodash(cbioportal-frontend-commons, react-mutation-mapper). But at least we are ready once it's removed from the dependency.

This is related to https://github.com/oncokb/oncokb-public/pull/1134


src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
+type DebouncedFunction<F extends (...args: any[]) => any> = (
```
zhx828, line 1: This file needs to be reviewed thoroughly. I tried to create some tests to cover each method in Lodashutils.spec.ts. But I might have missed some edge cases.

zhx828: I also hope to move this file to cbioportal-frontend repo so we can retire the lodash dependency from cbioportal-frontend-commons, react-mutation-mapper repos.

zhx828: Some methods are from https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore, I filled in the types. The rest are created by me.

src/main/webapp/app/pages/companionDiagnosticDevicesPage/companionDiagnosticDevicePage.tsx
```diff
@@ -507,14 +501,16 @@ const CompanionDiagnosticDevicePage: React.FunctionComponent<{}> = () => {
...
             placeholder={'Gene(s)'}
-            options={_.chain(parsedCDx)
-              .map(cdx => ({
-                label: cdx.biomarkerAssociation.gene,
-                value: cdx.biomarkerAssociation.gene,
-              }))
-              .uniqBy('label')
-              .sortBy('label')
-              .value()}
+            options={sortByKey(
```
zhx828, line 504: If possible, I'd like to do some chaining in the future too.

src/main/webapp/app/pages/genePage/FdaUtils.tsx
```diff
@@ -1,132 +1,3 @@
-import React from 'react';
-import _ from 'lodash';
-import { Linkout } from 'app/shared/links/Linkout';
-import { PMIDLink } from 'app/shared/links/PMIDLink';
-import { concatElements, concatElementsByComma } from 'app/shared/utils/Utils';
-import { DefaultTooltip } from 'cbioportal-frontend-commons';
-import * as styles from 'app/index.module.scss';
 import { ONCOKB_TM } from 'app/config/constants';
 
-const DrugLabel: React.FunctionComponent<{ drug: any }> = props => {
```
zhx828, line 10: The following methods are not used anywhere.

src/main/webapp/app/typings.d.ts
```diff
@@ -20,3 +20,5 @@ declare module '*.png' {
 declare module 'availity-reactstrap-validation';
 declare module 'react-responsive-tabs';
 declare module 'victory';
+declare module 'lodash.debounce';
+declare module 'lodash';
```
jfkonecn, line 24: Why is this needed?

tsconfig.json
```diff
@@ -1,14 +1,15 @@
...
-    "target": "es5",
+    "target": "es6",
     "module": "esnext",
     "moduleResolution": "node",
     "sourceMap": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "removeComments": false,
     "noImplicitAny": true,
+    "skipLibCheck": true,
```
jfkonecn, line 12: Just curious, why did you need to add this?

src/main/webapp/app/pages/companionDiagnosticDevicesPage/companionDiagnosticDevicePage.tsx
jfkonecn: You may be able to [do that more easily](https://github.com/tc39/proposal-pipeline-operator) in the future.

jfkonecn (changes requested):

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+/**
+ * Creates an array of elements split into two groups, the first of which contains elements `predicate` returns truthy for, the second of which contains elements `predicate` returns falsy for.
+ * @param array - The array to iterate over.
+ * @param predicate - The function invoked per iteration.
+ * @returns The array of grouped elements.
+ */
+export function partition<T>(array: T[], predicate: (datum: T) => boolean) {
+  return array.reduce(
+    ([pass, fail], elem) =>
+      predicate(elem) ? [[...pass, elem], fail] : [pass, [...fail, elem]],
```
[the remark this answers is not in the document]
calvinlu3: Yeah, I would use `Array.push` to modify in place instead of spreading the array on each iteration since that creates a new array.

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+ * @param arrays - The arrays to inspect.
+ * @returns The new array of filtered values.
+ */
+export function xor<K extends number | string>(...arrays: K[][]) {
+  // Manually flatten the arrays into a single array
+  const flattened = flatten(arrays);
+
+  // Create a frequency map to count occurrences of each element
+  const frequencyMap = flattened.reduce((acc, val: K) => {
+    acc[`${val}`] = (acc[`${val}`] || 0) + 1;
```
calvinlu3, line 256: If we have two arrays:
`[ '100', 2, 3] `
`[ 100, 2, 3] `
Using string formatting for the key will cause an issue here because `${100}` and `${'100'}` will both return `'100'`. I would expect xor will return `['100', 100]`?

calvinlu3 (changes requested):

src/main/webapp/app/shared/utils/LodashUtils.ts
zhx828: Interestingly I would expect it to fail within the usage in FormValidationUtils. The args are spread to the `debouncedCompanyNameValidator` which have different types. Did I misunderstand the use of `unknown`?

jfkonecn: `unknown` means you have no idea what the type is (so it could be a union type which is your case). If anything cares about the type then you have to prove it's the expected type with type checks.
`any` means turn off type checking.

zhx828: Doesn't unknown forbid unknown to be assigned to other type? https://stackoverflow.com/a/51439876/1868022

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+  };
+}
+
+/**
+ * Flattens a two-dimensional array into a one-dimensional array.
+ * @param array - The array to flatten.
+ * @returns The new flattened array.
+ */
+export function flatten<T>(array: T[][]) {
+  return array.reduce((a, b) => a.concat(b), []);
```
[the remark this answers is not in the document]
zhx828: Yup, I didn't include es2019 into typescript which I think I could. I will update.

src/main/webapp/app/shared/utils/LodashUtils.ts
jfkonecn: Let's talk about it on a call. I don't feel like I can explain it well in a comment.

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+  return find([...array].reverse(), predicate);
+}
+
+/**
+ * Creates an object composed of keys generated from the results of running each element of `array` through `key`.
+ * @param array - The array to iterate over.
+ * @param key - The key to group by.
+ * @returns The composed aggregate object.
+ */
+export function groupBy<T extends object & NonArrayObject, K extends keyof T>(
```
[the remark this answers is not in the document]
zhx828: Object.groupBy is only included in es2024. Our version of typescript supports up to 2020. We can have a separate pr to upgrade typescript.

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+  return undefined;
+}
+
+/**
+ * This method is like `find` except that it iterates over elements of `array` from right to left.
+ * @param array - The array to inspect.
+ * @param predicate - The function invoked per iteration.
+ * @returns The matched element, else `undefined`.
+ */
+export function findLast<T>(array: T[], predicate: (datum: T) => boolean) {
```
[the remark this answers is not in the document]
zhx828: Object.groupBy is only included in es2023. Our version of typescript supports up to 2020. We can have a separate pr to upgrade typescript.

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+  }, []);
+}
+
+/**
+ * Iterates over elements of `array`, returning the first element `predicate` returns truthy for.
+ * @param array - The array to inspect.
+ * @param predicate - The function invoked per iteration.
+ * @returns The matched element, else `undefined`.
+ */
+export function find<T>(array: T[], predicate: (datum: T) => boolean) {
```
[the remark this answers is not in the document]
zhx828: find is supported in es6. This method is not used anywhere. But let's keep it here. I do plan to move these to cbioportal. I will remove eventually if it's not used.

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Checks if `key` is a direct property of `object`.
+ * @param obj - The object to query.
+ * @param key - The path to check. The path can be nested prop. For instance: a.b.c
+ * @returns `true` if `key` exists, else `false`.
+ */
+export function has<T extends object & NonArrayObject>(obj: T, key: string) {
+  const keyParts = key.split('.');
+  let clone = JSON.parse(JSON.stringify(obj));
```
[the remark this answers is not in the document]
zhx828: good point.

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+export function intersection(...arrays: (number | string)[][]) {
+  return arrays.reduce((a, b) => a.filter(c => b.includes(c)));
+}
+
+/**
+ * Checks if `value` is classified as a `Number` primitive or object.
+ * @param value - The value to check.
+ * @returns `true` if `value` is a number, else `false`.
+ */
+export function isNumber(value: any) {
```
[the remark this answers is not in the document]
zhx828: Really cool

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+  return true;
+}
+
+/**
+ * Creates an array of unique values that are included in all given arrays.
+ * @param arrays - The arrays to inspect.
+ * @returns The new array of shared values.
+ */
+export function intersection(...arrays: (number | string)[][]) {
+  return arrays.reduce((a, b) => a.filter(c => b.includes(c)));
```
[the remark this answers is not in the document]
zhx828: Is the concern on dup within arrays, or dup within a or b?

src/main/webapp/app/shared/utils/LodashUtils.ts
```diff
@@ -0,0 +1,273 @@
...
+ * @returns The sorted array.
+ */
+export function sortBy<T extends object & NonArrayObject, K extends keyof T>(
+  array: T[],
+  comp: (datum: T) => number | string
+) {
+  return array.sort((a: T, b: T) => {
+    const ra = comp(a);
+    const rb = comp(b);
+    if (typeof ra === 'string' || typeof rb === 'string') {
```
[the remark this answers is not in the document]
zhx828: I don't want to complicate the case. All methods within this utils do not support mix type.

src/main/webapp/app/shared/utils/ReactTableUtils.ts
```diff
@@ -77,14 +77,14 @@ export function defaultFdaImplicationSortMethod(a: any, b: any): number {
 }
 
 export function sortNumber(a: number, b: number): number {
-  if (!_.isNumber(a)) {
-    if (!_.isNumber(b)) {
+  if (!isNumber(a)) {
```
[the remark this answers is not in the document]
zhx828: I will update the type of sortNumber. The usage of the method is to process API data which they can be undefined.

tsconfig.json
zhx828: Some of the dependencies use lodash and do not have proper types.

src/main/webapp/app/shared/utils/LodashUtils.ts
zhx828: Methods in this util do not support mix types.

jfkonecn (approved):

calvinlu3 (approved):
LGTM

PR #1146, opened 2024-08-27
Modify company data schema to include licensing details

jfkonecn:
closes oncokb/oncokb-pipeline#442

PR #1151, opened 2024-09-09
Added termination warning email button

jfkonecn:
oncokb/oncokb-pipeline#443

@zhx828 The email from field is being ignore. It uses the oncokb dev email as the sender. Is it okay to use that one or do we need to be able to customize the from email address? 

FYI this includes the [schema changes too](https://github.com/oncokb/oncokb-public/pull/1146)


src/main/java/org/mskcc/cbio/oncokb/web/rest/MailsController.java
```diff
@@ -117,4 +141,47 @@ public List<MailTypeInfo> getMailsTypes() {
         return mailTypeInfos;
     }
 
+    @GetMapping("mails/termination-warning/{id}")
```
zhx828, line 144:
```suggestion
    @GetMapping("mails/termination-warning/{companyId}")
```

src/main/java/org/mskcc/cbio/oncokb/web/rest/MailsController.java
```diff
@@ -117,4 +141,47 @@ public List<MailTypeInfo> getMailsTypes() {
...
+            ZonedDateTime zoned = date.atZone(ZoneId.of(Constants.NY_ZONE_ID));
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MMMM d, yyyy");
+            formattedDate = formatter.format(zoned);
+        } catch (NullPointerException e) {
+            throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Please save the termination date first.");
+        }
+
+        EmailAddresses addresses = this.applicationProperties.getEmailAddresses();
+
+        String from = addresses.getTechDevAddress();
```
zhx828, line 162: This should from license address

src/main/java/org/mskcc/cbio/oncokb/web/rest/MailsController.java
```diff
@@ -117,4 +141,47 @@ public List<MailTypeInfo> getMailsTypes() {
...
+        String cc = addresses.getLicenseAddress();
+        String bcc = users.stream().map(UserDTO::getEmail).collect(Collectors.joining(";"));
+        String subject = "Urgent: OncoKB API License Expiration on " + formattedDate + " - Action Required";
+        String content = "Dear " + company.getName() + ",\n\n" +
+                "We are writing to inform you that your API license for OncoKB is set to expire on " + formattedDate + ".\n\n" +
+                "To ensure uninterrupted access to the OncoKB database and its valuable resources, it is essential to initiate the renewal process as soon as possible.\n\n" +
+                "To renew the license, please reply to this email.\n\n" +
```
zhx828, line 169:
```suggestion
                "If you would like to continue using OncoKB, please let us know, and we will work with your company’s representative to facilitate the license renewal.\n\n" +
```

src/main/webapp/app/pages/companyPage/CompanyPage.tsx
```diff
@@ -113,8 +119,18 @@ enum SimpleConfirmModalType {
   NA,
   DELETE_COMPANY,
   UPDATE_COMPANY,
+  SEND_TERMINATION_NOTICE,
 }
 
+const defaultPayload: TerminationEmailDTO = {
+  subject: '',
+  from: 'admin@oncokb.org',
+  cc: 'licenses@oncokb.org',
```
zhx828, line 128: let's don't expose these addresses in the code if they are not required.

zhx828 (changes requested):
I will do a product review later today. Some quick notes added.

src/main/java/org/mskcc/cbio/oncokb/web/rest/MailsController.java
jfkonecn: When I do my local testing it seems to always send from the dev email.

zhx828: Probably because you are using the gmail account? That account only has one address. In production, we will be able to switch to other email addresses.

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
```diff
@@ -85,6 +86,20 @@ private UnknownMailTypeException() {
         }
     }
 
+    public void sendEmail(TerminationEmailDTO dto) throws MessagingException {
+        log.debug("Send email to company ID '{}' with subject '{}' and content={}",
+            dto.getCompanyId(),  dto.getSubject(), dto.getContent());
+
+        MimeMessage mimeMessage = javaMailSender.createMimeMessage();
+        MimeMessageHelper message = new MimeMessageHelper(mimeMessage, false, StandardCharsets.UTF_8.name());
+        message.setFrom(dto.getFrom());
```
zhx828, line 95: Let's add `message.setTo();` as well.

zhx828 (changes requested):
One minor change requested. Thanks!

src/main/java/org/mskcc/cbio/oncokb/service/MailService.java
jfkonecn: I'm using BCC instead of To.

zhx828: oh right. Thanks! Good to go.

PR #1176, opened 2024-10-21
Add germline core proxy config

zhx828:


src/main/java/org/mskcc/cbio/oncokb/service/ApiProxyService.java
```diff
@@ -27,7 +27,12 @@ public class ApiProxyService {
 
     public URI prepareURI(HttpServletRequest request) throws URISyntaxException {
         String queryString = request.getQueryString();
-        return new URI(applicationProperties.getApiProxyUrl() + request.getRequestURI() + (queryString == null ? "" : "?" + queryString));
+        String defaultApiProxyUrl = applicationProperties.getApiProxyUrl();
+        String germlineParam = request.getParameter("germline");
```
jfkonecn, line 31: Is there a way we could make a new base endpoint altogether? I feel like an annoying bug will happen because of this.

jfkonecn (changes requested):

zhx828:
Try to add an auth requirement on top.

jfkonecn (approved):

PR #1187, opened 2024-11-04
Updates based on the UX feedback

zhx828:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/601


src/main/webapp/app/shared/nav/StickyMiniNavBar.tsx
```diff
@@ -161,11 +161,11 @@ export default function StickyMiniNavBar({
...
 
   return (
-    <Container
+    <div
       className={classNames(
-        'container',
         styles.container,
-        isSticky ? styles.containerSticky : ''
+        isSticky ? styles.containerSticky : '',
+        'container'
```
calvinlu3, line 168: Whats the difference between the class and the component?

src/main/webapp/app/shared/icons/ShowHideToggleIcon.tsx
calvinlu3, line 1: The icon inside the circle doesn't seem to be centered. Seems to be slightly leaning left. Do you see that too?

src/main/webapp/app/shared/nav/StickyMiniNavBar.tsx
zhx828: Good question. Initially I removed the container, then added back the class. Let me try to revert to the component.

src/main/webapp/app/shared/icons/ShowHideToggleIcon.tsx
zhx828: Oh yeah, it might relate to some changes after introducing the icon. I will look inot.

zhx828: Not sure why I can't reproduce locally.

calvinlu3: I was looking at the screenshots, but locally it is fine for me too.

src/main/webapp/app/shared/nav/StickyMiniNavBar.tsx
```diff
@@ -161,11 +161,11 @@ export default function StickyMiniNavBar({
   )[0]?.id;
 
   return (
-    <div
+    <Container
       className={classNames(
+        'container',
```
calvinlu3: Can remove class now

calvinlu3 (approved):
LGTM. Screenshots still failing.

PR #1200, opened 2025-01-29
v4.25

jfkonecn:


src/main/webapp/app/config/constants.tsx
```diff
@@ -866,6 +866,7 @@ export type DataRelease = {
 };
 
 export const DATA_RELEASES: DataRelease[] = [
+  { date: '03012025', version: 'v4.25' },
```
calvinlu3, line 869:
```suggestion
  { date: '01302025', version: 'v4.25' },
```

calvinlu3 (changes requested):
Side note: Your company PR changes are merged to master. Can you go to beta website (https://beta.oncokb.org/companies/93) and verify if your feature is good to release?

We will also need to do a DB migration

src/main/webapp/app/config/constants.tsx
jfkonecn: don't have dyslexia

PR #1203, opened 2025-02-18
v4.26

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/663

[scripts/generate-news-sections.js](https://github.com/oncokb/oncokb-public/pull/1203/files#diff-5c105efc947ef2c431be6802dac62ee5515c076d59cd4b726c2c6d96b08b28fa) reads all markdown files and then code generates a react component for each one. Most of the code in the script is to adjust the DOM so that the correct react components are used.


calvinlu3 (changes requested):
Nice work on the news page script - way easier!

Some changes stilled needed:
1. New summary image and update index.html
2. Fix ss test

src/main/webapp/app/pages/newsPage/code-generated/NewsContent022025.tsx
```diff
@@ -0,0 +1,209 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import {
+  AlterationPageLink,
+  getAlternativeGenePageLinks,
+} from 'app/shared/utils/UrlUtils';
+import { NewlyAddedGenesListItem } from 'app/pages/newsPage/NewlyAddedGenesListItem';
+
+export default function NewsContent022025() {
```
calvinlu3, line 9: We may have multiple releases in a month, so may need to add a date in future

calvinlu3 (approved):

PR #1209, opened 2025-03-01
Ignore TokenExpiredException in Sentry

calvinlu3:
Resolves https://github.com/oncokb/oncokb/issues/3866

No need to catch this exception in Sentry. It is for the user


bprize15:
> Resolves [oncokb/oncokb#3866](https://github.com/oncokb/oncokb/issues/3866)
> 
> No need to catch this exception in Sentry. It is for the user

Could this be useful information that we want to keep track of and report to the user? Ideally, the user should just read the error message, but could it be useful in case the user emails us asking for help?

calvinlu3:
We intentionally ignore non criticl exceptions like `BadCredentialsException` and `EmailAlreadyUsedException` already to reduce noise in our Sentry logs as these are not system failures.
If tracking non critical exceptions is useful for support, then maybe there is a way to categorize exception priority in Sentry (not sure, seems like a valid feature)? Until we see a use case for tracking these, I will just ignore for now?

bprize15 (approved):
LGTM

PR #1212, opened 2025-03-07
Fix/multiple pub web token

calvinlu3:
Fixes https://github.com/oncokb/oncokb/issues/3865


jfkonecn (commented):
So is someone using our api when we're in the middle of creating a new token for them?

calvinlu3:
> So is someone using our api when we're in the middle of create a new token for them?

@jfkonecn Let me look into this issue again. I may have mistakenly thought that the daily cronjob creates a new pub web token if its expired.

calvinlu3:
Going to close the PR for now because the issue is not reproducible and will revisit when this issue occurs again. The last and only time we saw this exception was in January. The oncokb_public database only has one public website token.

PR #1232, opened 2025-08-18
Added request url/response code logging.

jfkonecn:


jfkonecn:
@calvinlu3 I removed the debug logic to reduce the risk that someone would log sensitive information and the performance hit if someone set logging to debug mode.
Let me know what you think.

calvinlu3 (approved):

PR #1240, opened 2025-09-02
add gene type badge to variants when searching

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/856


src/main/webapp/app/components/searchOption/SearchOption.tsx
```diff
@@ -26,6 +29,23 @@ export enum SearchOptionType {
...
+      geneticType.toLowerCase() as GENETIC_TYPE
+    ) ? (
+    <div>
+      <GeneticTypeTag
+        isGermline={geneticType === 'GERMLINE'}
+        className={'ml-2'}
+      />
+    </div>
+  ) : (
+    <></>
```
jfkonecn, line 45: If we end up here, then does that mean our data is corrupted?

jfkonecn (changes requested):

bprize15:
Since this PR relies on a core change, and we can't merge core, we can't merge this

src/main/webapp/app/components/searchOption/SearchOption.tsx
bprize15: Yes it does, I would never expect this to happen

PR #1303, opened 2026-01-28
Add germline specific annotation endpoints

calvinlu3:
resolves https://github.com/oncokb/oncokb-pipeline/issues/1076


jfkonecn (approved):

jfkonecn (approved):

bprize15 (approved):
LGTM

jfkonecn (approved):

src/main/webapp/app/components/searchOption/SearchOption.tsx
```diff
@@ -52,6 +57,17 @@ type SearchOptionProps = {
...
 };
+
+function getSearchResultIconOncogenicity(data: ExtendedTypeaheadSearchResp) {
+  if (data.pathogenicity === PATHOGENICITY.PATHOGENIC) {
+    return ONCOGENICITY.ONCOGENIC;
+  }
+  if (data.pathogenicity === PATHOGENICITY.LIKELY_PATHOGENIC) {
+    return ONCOGENICITY.LIKELY_ONCOGENIC;
+  }
+  return data.oncogenicity;
```
bprize15, line 68: Do we not have icons for benign, likely benign, etc?

src/main/webapp/app/shared/tiles/tile-utils.tsx
```diff
@@ -322,6 +325,17 @@ export function SomaticGermlineAlterationTiles({
     GENOME_NEXUS_ANNOTATION_BASE_URL,
   ]);
 
+  if (isPathogenicVariantUmbrella) {
+    tiles.push(
+      createHighestLevelOfEvidenceTileProps(
```
bprize15, line 330: I assume curators are never setting penetrance for pathogenic variants?

bprize15 (approved):
Just a couple questions, otherwise LGTM

src/main/webapp/app/components/searchOption/SearchOption.tsx
calvinlu3: Good point, we don't have those icons, but I wonder if they can be mapped to the somatic "neutral" icons. Thanks for pointing that out

src/main/webapp/app/shared/tiles/tile-utils.tsx
calvinlu3: There is penetrance for `Pathogenic Variants`, but curator decided we won't show that tile. We will only show highest level tile.

PR #1334, opened 2026-04-16
Added Quick User View Modal

jfkonecn:
[Per Moriah](https://oncokb.slack.com/archives/C1BLD9GSV/p1776277547420109)

closes https://github.com/oncokb/oncokb-pipeline/issues/1160


src/main/webapp/app/pages/userManagement/UserDetailsPage.tsx
```diff
@@ -491,6 +541,12 @@ export default class UserDetailsPage extends React.Component<{
           onCancel={() => this.cancelUpdateActiveStatus()}
           onConfirm={this.updateActiveStatus.bind(this)}
         />
+        <UserQuickViewModal
+          show={this.showQuickViewModal}
+          user={this.currentSelected.user}
+          onClose={() => this.closeQuickViewModal()}
+          onUpdateActiveStatus={this.updateQuickViewUserActiveStatus.bind(this)}
```
calvinlu3: Adding `@action.bound` would be simpler

src/main/webapp/app/pages/userManagement/UserQuickViewModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+  if (isRequestingApiAccess && !authorities.includes(AUTHORITIES.API)) {
+    authorities.push(AUTHORITIES.API);
+  }
+
+  return authorities;
+}
+
+function getUserPageLink(login: string) {
+  return `${window.location.origin}/users/${login}`;
```
calvinlu3: Might be better to take advantage of react router and use 
```
<Link to={`/users/${login}`} />
```
instead of ExternalLink to avoid full page reloads and since we're still routing within our own app.

calvinlu3 (changes requested):

src/main/webapp/app/pages/userManagement/UserQuickViewModal.tsx
jfkonecn: I want to allow people to be able to open the page in a new tab, but you make a fair point. I'll change it. People can right click to open on a new tab.

calvinlu3 (approved):
LGTM

PR #1336, opened 2026-04-16
Add content transfer encoding to emails and add manual key verification

calvinlu3:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/1156
  
### Changes
- Content Transferm Encoding: Sets base64 encoding on outgoing emails to prevent link corruption in email clients that mangle special characters in URLs.
- Remove `login` and `hasGracePeriod` from activation URL. These can be figured out based off of activationKey alone
- Visiting `/account/verify` without a key param now shows a form where users can paste their activation key manually, as a fallback when email links don't work.
- Fix message when user signs up with personal email (was incorrectly saying they have access)

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
@@ -156,6 +156,7 @@ public class AccountResource {
         UserDetailsDTO existingUserDetails = userDetailsService.findOneByUser(existingUser)
             .orElseThrow(() -> new CustomMessageRuntimeException("User details could not be found"));
         boolean hasGracePeriod = AccountRequestStatus.PENDING.equals(existingUserDetails.getAccountRequestStatus())
+            && existingUserDetails.getLicenseType() != null
             && SecurityUtils.isWithinActivationGracePeriod(existingUser, existingUserDetails.getLicenseType());
         boolean newUserActivation = !existingUser.getActivated();
 
--- a/src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
@@ -5,6 +5,7 @@ import org.mskcc.cbio.oncokb.domain.Company;
 import org.mskcc.cbio.oncokb.domain.CompanyCandidate;
 import org.mskcc.cbio.oncokb.domain.Token;
 import org.mskcc.cbio.oncokb.domain.User;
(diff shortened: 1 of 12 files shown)


src/main/webapp/app/pages/companyPage/CompanyPage.tsx
```diff
@@ -381,13 +381,12 @@ export default class CompanyPage extends React.Component<ICompanyPage> {
...
     try {
-      const activated = await client.activateAccountUsingGET({
+      const result = await client.activateAccountUsingGET({
         key: user.activationKey,
-        login: user.login,
       });
       const updatedUser = {
         ...user,
-        activated,
+        activated: result.activated,
```
jfkonecn, line 389: Is this code currently broken in production?

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
```diff
@@ -145,46 +146,44 @@ public GracePeriodBlacklistVM getGracePeriodBlacklist() {
...
+        }
 
-            boolean newUserActivation = !userOptional.get().getActivated();
-            userOptional = userService.activateRegistration(key);
+        User existingUser = userOptional.get();
+        UserDetailsDTO existingUserDetails = userDetailsService.findOneByUser(existingUser)
+            .orElseThrow(() -> new CustomMessageRuntimeException("User details could not be found"));
+        boolean hasGracePeriod = AccountRequestStatus.PENDING.equals(existingUserDetails.getAccountRequestStatus())
+            && SecurityUtils.isWithinActivationGracePeriod(existingUser, existingUserDetails.getLicenseType());
+        boolean newUserActivation = !existingUser.getActivated();
```
jfkonecn, line 161: If the user isn't active it doesn't mean they're new. They could be an old user that got deactivated or was never activated.

jfkonecn (changes requested):

jfkonecn (changes requested):
see comments

src/main/java/org/mskcc/cbio/oncokb/web/rest/AccountResource.java
calvinlu3: I think its confusing because this endpoint is called `activate`, but it really is just verifying the email.
The `activated` field from User model means whether the user has been approved or not right? 

If they got disapproved or never got approved, then `user.activated = false` and if they proceed to verify their email address in this endpoint, then they are treated as a new activation.

This is the same logic as the original.

jfkonecn: alright, leave it for now then

jfkonecn (approved):

PR #1337, opened 2026-04-17
Split Slack license clarification emails into hospital and commercial flows

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/1164

AI Generated Summary:

This PR splits the old single Slack “license options” email into two separate clarification flows: one for hospital use and one for commercial use. Each now appears as its own automatic email option in Slack, with its own mail type, action IDs, block IDs, and default modal content.

It also adds the new hospital and commercial email templates from the ticket, routes both through the license email sender/CC behavior, and removes the remaining old “license options” naming so the subject generation is consistent around email.license.clarify.title. The website registration flow is unchanged.


scripts/register-random-user.sh
calvinlu3, line 1: Did you intend to commit this?

calvinlu3 (approved):

scripts/register-random-user.sh
jfkonecn: yes, now we don't have to go to the registration page to make a new user.

PR #1341, opened 2026-04-24
Handle invalid variant annotations in the client

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/1152

  The main change is a new normalized annotation-result layer in `src/main/webapp/app/store/AnnotationResult.ts` and `src/main/webapp/app/store/` AnnotationStore.ts. The store now:

  - exposes a discriminated valid/invalid annotation union
  - flags malformed single-alteration queries like `L768S,` as invalid when variantExist is false


src/main/webapp/app/store/AnnotationResult.ts
```diff
@@ -0,0 +1,45 @@
...
+          drugs: treatment.drugs ?? [],
+        })),
+      })),
+    })),
+  };
+}
+
+export function isInvalidVariantAnnotation(annotation: VariantAnnotation) {
+  return (
+    !annotation.variantExist && annotation.oncogenic === ONCOGENICITY.UNKNOWN
```
jfkonecn: Please carefully double check this logic.

jfkonecn:
It doesn't look like there's a difference in the response for a bad query in the variant API

PR #1351, opened 2026-04-30
Hospitals now have to justify API access

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/1187

### DIFF
--- a/src/main/webapp/app/components/newAccountForm/ApiAccessSection.tsx
+++ b/src/main/webapp/app/components/newAccountForm/ApiAccessSection.tsx
@@ -47,9 +47,10 @@ export default function ApiAccessSection({
             <p>
               Would you like programmatic access to the {ONCOKB_TM} database via
               our API? API access allows a user to simultaneously annotate
-              multiple tumor mutations with {ONCOKB_TM} data and provides a text
-              file output. {ONCOKB_TM} API access may also enable the user to
-              leverage {ONCOKB_TM} alongside other platform APIs.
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/components/newAccountForm/ApiAccessSection.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <h5>API Access</h5>
+      </Col>
+      <Col md="9">
+        {licenseType === LicenseType.ACADEMIC && (
+          <>
+            <p>
+              Would you like programmatic access to the {ONCOKB_TM} database via
+              our API? API access allows a user to simultaneously annotate
+              multiple tumor mutations with {ONCOKB_TM} data and provides a text
+              file output. {ONCOKB_TM} API access may also enable the user to
```
bprize15: Is "provides a text file output" accurate?

bprize15 (approved):
One small comment, otherwise LGTM

src/main/webapp/app/components/newAccountForm/ApiAccessSection.tsx
jfkonecn: @calvinlu3 ?

calvinlu3: Yeah I think we want to either omit the text file output or state in another way. 
Maybe you want to make it clear that they will need api access to use our oncokb-annotator? But they can also just use the api as well.

PR #1359, opened 2026-05-15
Fixed slack error and removed sentry

jfkonecn:
# Summary

Sentry is no longer part of the supported monitoring stack, so this PR removes the remaining Sentry integration from the application.

While doing that cleanup, this also addresses a Slack notification issue where oversized payload content could cause delivery problems. Previously, that kind of failure was easier to spot through Sentry reporting. Without Sentry in place, the same class of issue is harder to diagnose unless the application logs the failure clearly. This change replaces that reporting path with application logging and adds truncation coverage for the oversized Slack duplicate-user message case.

## What changed

- Removed Sentry from the backend and frontend
- Replaced Sentry-based Slack failure reporting with standard application logs
- Added coverage for the Slack large-text case so oversized duplicate-user warning content is truncated before sending


src/main/webapp/app/index.tsx
```diff
@@ -115,25 +114,14 @@ superagent.Request.prototype.end = function (callback) {
     const currentCount = Number(localStorage.getItem(STORAGE_KEY));
 
     // If we've already reloaded more than the threshold for this hour,
-    // log to Sentry (if configured) or show a generic error to prevent
-    // infinite reload.
+    // log locally and show a generic error to prevent infinite reload.
     if (currentCount > WEBSITE_RELOAD_TIMES_THRESHOLD) {
-      let message =
+      console.error(
```
calvinlu3, line 119: Does console.error get picked up by datadog?

calvinlu3 (commented):
You have a couple of failing IT

src/main/webapp/app/index.tsx
jfkonecn: No, we'll need to add a data dog npm package, but I believe we need to upgrade react last time I checked.

calvinlu3 (approved):

PR #1377, opened 2026-06-02
add user email sent to user table and quick view

ReshmaRamaiah10:
https://github.com/oncokb/oncokb-pipeline/issues/1201


src/main/java/org/mskcc/cbio/oncokb/domain/enumeration/MailType.java
```diff
@@ -94,6 +94,10 @@ public enum MailType {
         .description("Clarify - Commercial use")
         .titleKey("email.license.clarify.title")
         .stringTemplateName("clarifyCommercialUseString.txt"))
+    , LICENSE_OPTIONS(new MailTypeBuilder()
```
jfkonecn: add a comment

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -572,14 +572,23 @@ public void changePassword(String currentClearTextPassword, String newPassword)
...
             })
             .collect(Collectors.toList());
+
+        // Batch-load all mails and populate each UserDTO
+        Map<Long, List<org.mskcc.cbio.oncokb.service.dto.UserMailsDTO>> mailsByUserId =
+            userMailsService.findUserMailsGroupedByUserId(users);
+        userDTOs.forEach(dto -> dto.setUserMails(
+            mailsByUserId.getOrDefault(dto.getId(), Collections.emptyList())
+        ));
+
```
jfkonecn: convert into one database

calvinlu3 (changes requested):
1. Adding userMails to UserDTO makes the endpoint load and serialize each user’s full mail history as part of the user list response. I feel like the endpoint should be lean and should not overfetch. 
We are trying to get everything about a user in one API call, which means its going to take more time/work to get the UserDTOs now since the dto is carrying more data. That means the UI needs to wait even though someone might only want to click the "Activated" button (for example), which doesn't require the mail.
2. The list of users is cached and you just added userMails to the userDTO. If a mail was sent, the user cache does not have the stale user evicted, so on frontend you might have stale user mails list.

I don't think we should add to the users endpoint, but just call a seperate endpoint to get userMails. It seems like you would only need to actually get the userMails when someone clicks on the quick view.

jfkonecn:
@ReshmaRamaiah10 let's walkthrough this PR during our meeting today.

jfkonecn (approved):

calvinlu3 (approved):

PR #1378, opened 2026-06-11
Update/code gen

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/1050

### `Highestsensitivelevel` and other `Highest*Level` are nullable, but swagger doesn't generate null.

For this task I modified `scripts/generate-api.js` to look for `(nullable)` in the description since we need to have upgrade our swagger to represent nulls

### There is a TumorType and TumorTypeEntity model that the API can return. We should consolidate the core to just return only the TumorType (apiModel) instead of TumorTypeEntity (database model).

I had to give up on this task. There are a few database models being returned in the API response. I think we just need to wait for the rewrite.

### DIFF
--- a/src/main/webapp/app/config/constants.tsx
+++ b/src/main/webapp/app/config/constants.tsx
@@ -591,7 +591,6 @@ export const DEFAULT_GERMLINE_ANNOTATION: GermlineVariantAnnotation = {
   geneSummary: '',
   genomicIndicators: [],
   highestDiagnosticImplicationLevel: 'NO',
-  highestFdaLevel: 'NO',
   highestPrognosticImplicationLevel: 'NO',
   highestResistanceLevel: 'NO',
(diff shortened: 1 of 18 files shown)


jfkonecn:
I just realized that I didn't do all the tasks in the issue. I just did the code generation.

src/main/webapp/app/config/constants.tsx
```diff
@@ -497,9 +499,9 @@ export const DEFAULT_GERMLINE_VARIANT = {
   penetranceDescription: '',
 };
 
-export const DEFAULT_QUERY = {
+export const DEFAULT_QUERY: Query = {
```
jfkonecn, line 502: `normalizedAlteration` is not on the query object.

calvinlu3 (approved):

PR #1389, opened 2026-07-02
Improve message when activation key is already used.

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/1203
image


jfkonecn (approved):
Thanks for taking care of this!

PR #1399, opened 2026-07-30
Feat/add content news

bprize15:


src/main/java/org/mskcc/cbio/oncokb/web/rest/NewsResource.java
```diff
@@ -0,0 +1,143 @@
...
+    public NewsResource(ObjectMapper objectMapper, ApplicationProperties applicationProperties) {
+        this.objectMapper = objectMapper;
+        this.applicationProperties = applicationProperties;
+    }
+
+    private static volatile List<ContentNews> contentNews = null;
+
+    @PostConstruct
+    public void init() { // fetch content news from GitHub
+        String token = applicationProperties.getOncokbDataToken();
```
jfkonecn, line 49: So currently we're using the API proxy service to do downloads. I really don't like the idea of having to make changes to core for this change, but maybe we should to be consistent?

jfkonecn (approved):

PR #1409, opened 2026-09-02
Fix: Added paging to User Details page

jfkonecn:
## Summary
Adds server-side paging and filtering to the User Details page to reduce load time from about 40 seconds to under a second on production-sized data.

## Problem
The User Details page loaded all users at once, causing very slow initial load and interactions.

## Root cause
The page relied on large unpaged user fetches and did filtering client-side, which did not scale with production data size.

## Fix approach
- Switch User Details retrieval to paginated backend requests.
- Add request payload support for filters (q, emailVerified, licenseTypes, roles) on /api/users/registered.
- Query paged user IDs first, then hydrate only users for the current page.
- Update frontend table to manual pagination, server-side search, and role filtering.
- Regenerate API client and apply required call-site updates.

Validated manually against production database data:
- Open User Details page and verify first page loads quickly (target under 1 second).
- Verify query filtering works for login, email, first name, last name, job title, company name, city, and country.
- Verify role filtering works and excludes service-account role from selectable options.
- Verify email verification and license-type filtering continue to work.
- Verify navigating to next page returns correct results.

## Risk / Rollout
- Risk is primarily around pagination/filter correctness and API contract changes for /api/users/registered.
- Rollout is standard deployment; monitor user-management page behavior and filter and pagination results post-release.

## Related issues
Refs oncokb/oncokb-pipeline#1261


src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -619,6 +619,80 @@ public Page<UserDTO> getAllManagedUsers(Pageable pageable) {
...
+            return new PageImpl<>(Collections.emptyList(), pageable, userIdPage.getTotalElements());
+        }
+
+        Map<Long, UserDetails> userDetailsByUserId = userDetailsRepository.findByUserIdIn(userIds)
+            .stream()
+            .filter(userDetails -> userDetails.getUser() != null)
+            .collect(Collectors.toMap(userDetails -> userDetails.getUser().getId(), userDetails -> userDetails, (left, right) -> left));
+
+        List<UserDTO> userDTOs = new ArrayList<>(userIds.size());
+        for (User user : userRepository.findAllWithAuthoritiesAndUserMailsByIdIn(userIds)) {
```
jfkonecn: [ai-generated] Potential pagination correctness bug: `findAllWithAuthoritiesAndUserMailsByIdIn(userIds)` does not guarantee result ordering, so the returned page can come back in a different order than the `userIds` page query. That can cause records to shift positions between requests even with deterministic SQL ordering. Please reorder hydrated users by the original `userIds` sequence before creating `PageImpl`.

src/main/webapp/app/pages/userManagement/UserDetailsPage.tsx
```diff
@@ -532,6 +602,26 @@ export default class UserDetailsPage extends React.Component<{
   ];
 
   render() {
+    const tableFilters: React.FunctionComponent = () => (
+      <div style={{ minWidth: '300px' }}>
+        <Select
```
jfkonecn, line 610: [ai-generated] Accessibility follow-up: this `react-select` control is only identified by placeholder text, which is not a reliable accessible label. Please provide an explicit label (`<label htmlFor=...>`) or at least `aria-label`/`inputId` so screen readers can announce the role filter purpose consistently.

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
```diff
@@ -619,6 +619,80 @@ public Page<UserDTO> getAllManagedUsers(Pageable pageable) {
         return new PageImpl<>(userDTOs, pageable, users.size());
     }
 
+    @Transactional(readOnly = true)
+    public Page<UserDTO> getManagedUsersForUserDetailsPage(
```
jfkonecn, line 623: [ai-generated] Test coverage gap: this path introduces new server-side filtering dimensions (`q`, `emailVerified`, `licenseTypes`, `roles`) plus paged hydration in `getManagedUsersForUserDetailsPage`, but there is no targeted unit/integration coverage for filter semantics and stable paging behavior. Please add tests that at least verify role filtering, combined filter behavior, empty-filter path, and pagination order consistency across pages.

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
jfkonecn: [ai-generated] Fixed in `UserService#getManagedUsersForUserDetailsPage`: hydrated users are now indexed by id and emitted in the original `userIds` page order before creating `PageImpl`, preserving stable page ordering.

src/main/webapp/app/pages/userManagement/UserDetailsPage.tsx
jfkonecn: Fixed

src/main/java/org/mskcc/cbio/oncokb/service/UserService.java
jfkonecn: Added tests

src/main/java/org/mskcc/cbio/oncokb/config/Swagger2Config.java
calvinlu3, line 1: Seems unrelated to the issue, so just wondering what the reason and effect of this change is.

src/main/java/org/mskcc/cbio/oncokb/repository/UserRepository.java
```diff
@@ -48,6 +48,73 @@ public interface UserRepository extends JpaRepository<User, Long> {
...
+        value = "SELECT u.id " +
+            "FROM jhi_user u " +
+            "LEFT JOIN user_details ud ON ud.user_id = u.id " +
+            "WHERE u.login <> :anonymousLogin " +
+            "AND NOT EXISTS (" +
+            "  SELECT 1 FROM jhi_user_authority ua " +
+            "  WHERE ua.user_id = u.id AND ua.authority_name = :excludedAuthority" +
+            ") " +
+            "AND (:query IS NULL OR :query = '' " +
+            "  OR LOWER(u.login) LIKE CONCAT('%', LOWER(:query), '%') " +
```
calvinlu3, line 61: Searching still work well with all of these LIKEs and LOWERs?

src/main/java/org/mskcc/cbio/oncokb/web/rest/vm/RegisteredUsersRequestVM.java
```diff
@@ -0,0 +1,44 @@
+package org.mskcc.cbio.oncokb.web.rest.vm;
+
+import org.mskcc.cbio.oncokb.domain.enumeration.LicenseType;
+
+import java.util.List;
+
+public class RegisteredUsersRequestVM {
+    private String q;
```
calvinlu3: Rename query? I think the ai review tool should be able to nitpick this since its not a well named variable.

calvinlu3 (changes requested):

src/main/java/org/mskcc/cbio/oncokb/config/Swagger2Config.java
jfkonecn: I forgot to comment about that....
The AI skill should have caught this as out of scope...

I got annoyed having to undo metadata changes whenever I ran `yarn run fetchAPI`

src/main/java/org/mskcc/cbio/oncokb/repository/UserRepository.java
jfkonecn: Your point is well taken. I don't think it'll scale for a huge data set, but it seems to be working well enough at the moment.
https://beta.oncokb.org/admin/user-details

calvinlu3 (approved):

PR #1410, opened 2026-09-03
Support HGNC :: fusion nomenclature

calvinlu3:
Frontend support for the HGNC fusion nomenclature change, where fusion gene partners are separated with `::` (e.g. `BCR::ABL1 Fusion`) instead of a single hyphen.

### Canonical URL redirect

The API canonicalizes fusion names, so a page reached through the legacy hyphen form now settles on the canonical URL. `getCanonicalFusionAlteration` compares the alteration that was queried against the one the annotation came back with, and the alteration and cancer type pages `history.replace` to the canonical page link when they differ.

This is deliberately limited to fusions. The API also normalizes protein changes (`Val600Glu` to `V600E`), which the pages render as-is by design, so the helper only redirects when the annotated name contains `::`. Ambiguous queries such as `NKX3-1-ABL1 Fusion` are rejected by the API rather than normalized, so there is nothing to redirect to and the page is left alone.

Reactions in both pages are now collected and disposed in `componentWillUnmount`. This matters for the redirect specifically: the annotation can resolve after the user has navigated away, and an undisposed reaction would pull them back to the alteration page.

### Other changes

- `FUSION_SEPARATOR` constant, with a note on why the hyphen form is ambiguous (partner symbols like `NKX3-1` and `HLA-A` contain one).
- Fusion example row in the search info icon overlay.
- Regenerated API clients. `SomaticVariantAnnotation.alterationValidationError` is replaced by an `errors: ValidationError[]` array, which adds the `AMBIGUOUS_FUSION_SEPARATOR` type; `DEFAULT_ANNOTATION` follows. The regen also picks up the unrelated `/utils/genomicIndicators` endpoint and `inheritanceMechanisms` filters.

### Test plan

- `FusionUtils.spec.ts` covers the helper: hyphen queries that get normalized, queries that are already canonical, normalized protein changes and categorical alterations that must be left alone, the ambiguous fusion case, and missing values.
- Manually: `/gene/ABL1/somatic/BCR-ABL1 Fusion` redirects to the `::` form with search and hash preserved; `/gene/BRAF/somatic/Val600Glu` does not redirect.

### DIFF
--- a/src/main/webapp/app/shared/api/generated/OncoKbAPI.ts
+++ b/src/main/webapp/app/shared/api/generated/OncoKbAPI.ts
@@ -219,6 +219,12 @@ export type AnnotateStructuralVariantQuery = {
         'tumorType': string
 
 };
+export type ValidationError = {
+    'message': string
+
+        'type': "REFERENCE_ALLELE_MISMATCH" | "POSITION_OUT_OF_RANGE" | "REVERSED_POSITION_RANGE" | "MALFORMED_ALTERATION" | "AMBIGUOUS_FUSION_SEPARATOR"
+
+};
 export type ActionableGene = {
     'abstracts': string
 
@@ -279,6 +285,8 @@ export type AnnotateMutationByHGVScQuery = {
         'hgvsc': string
 
         'id': string
+
+        'inheritanceMechanisms': Array < "AUTOSOMAL_DOMINANT" | "AUTOSOMAL_RECESSIVE" | "X_LINKED_RECESSIVE" | "CARRIER" >
 
         'referenceGenome': "GRCh37" | "GRCh38"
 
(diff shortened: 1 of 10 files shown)


src/main/webapp/app/shared/api/generated/OncoKbAPI.ts
```diff
@@ -356,12 +364,6 @@ export type AnnotateCopyNumberAlterationQuery = {
 
         'tumorType': string
 
-};
-export type AlterationValidationError = {
-    'message': string
-
-        'type': "REFERENCE_ALLELE_MISMATCH" | "POSITION_OUT_OF_RANGE" | "REVERSED_POSITION_RANGE" | "MALFORMED_ALTERATION"
-
```
jfkonecn: Is this going away?

jfkonecn (approved):

PR #1411, opened 2026-09-09
Fix: Optimize Slack expand duplicate-user lookup performance

jfkonecn:
## Summary
Optimizes the Slack registration-request expand path by replacing duplicate-user N+1 lookups with batched and minimal-data queries, and refactoring duplicate result handling to a dedicated summary model.

## Problem
Expanding collapsed registration requests in Slack was slow, with performance degraded by repeated duplicate-user and rejection-mail checks.

## Root cause
The expand flow performed non-batched lookups and fetched broader user data than needed, causing avoidable query overhead in the duplicate-detection path.

## Fix approach
- Added native SQL duplicate-candidate retrieval in `UserRepository` with a projection that selects only required fields.
- Added batched native SQL lookup in `UserMailsRepository` to fetch user IDs that already received rejection emails.
- Added `UserMailsService` helper returning a `Set<Long>` for batched rejection status checks.
- Refactored Slack duplicate rendering to prefetch rejection states once and avoid per-user mail lookups.
- Introduced `PotentialDuplicateUserSummary` and updated `UserService` duplicate APIs to return this summary instead of reusing `UserDTO`.
- Updated `UserServiceIT` and added projection test helper for duplicate-candidate assertions.

- `./mvnw -Denforcer.skip=true -DskipTests compile` ✅
- `./mvnw -Denforcer.skip=true -Dtest=UserServiceIT#assertThatDuplicateUsersAreCaught test` ✅ (run with Java 8)

## Risk / Rollout
- Low functional risk: behavior is focused on duplicate detection and Slack expand rendering paths.
- Main change is query strategy (batched/minimal projection) rather than user-facing business rules.

## Related issues
Fixes oncokb/oncokb-pipeline#1262

### DIFF
--- a/src/main/java/org/mskcc/cbio/oncokb/repository/projection/PotentialDuplicateUserProjection.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/repository/projection/PotentialDuplicateUserProjection.java
@@ -1,0 +1,17 @@
+package org.mskcc.cbio.oncokb.repository.projection;
+
+public interface PotentialDuplicateUserProjection {
+    Long getId();
+
+    String getFirstName();
+
+    String getLastName();
+
+    String getEmail();
+
+    String getCompanyName();
+
+    String getCity();
+
+    String getCountry();
+}
--- a/src/main/java/org/mskcc/cbio/oncokb/repository/UserMailsRepository.java
+++ b/src/main/java/org/mskcc/cbio/oncokb/repository/UserMailsRepository.java
@@ -41,5 +41,11 @@ public interface UserMailsRepository extends JpaRepository<UserMails, Long> {
 
     List<UserMails> findUserMailByUserAndMailTypeIn(User user, List<MailType> mailTypes);
 
+    @Query(
+        value = "select distinct um.user_id from user_mails um where um.user_id in (:userIds) and um.mail_type in (:mailTypes)",
+        nativeQuery = true
+    )
+    List<Long> findUserIdsWithMailTypeIn(@Param("userIds") List<Long> userIds, @Param("mailTypes") List<String> mailTypes);
+
     void deleteAllByUser(User user);
 }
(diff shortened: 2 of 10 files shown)


jfkonecn (commented):
[ai-generated] AI code review completed for this diff. No merge blockers found across query safety, logging, accessibility impact, or test updates. Non-blocking follow-up: consider adding a regression test for the duplicate-candidate query behavior when  is null, since  in native SQL can evaluate to unknown and filter out all rows depending on invocation context.

jfkonecn (commented):
[ai-generated] AI code review completed for this diff. No merge blockers found across query safety, logging, accessibility impact, or test updates. Non-blocking follow-up: consider adding a regression test for the duplicate-candidate query behavior when userId is null, since the native SQL predicate "u.id <> :userId" can evaluate to unknown and filter out all rows depending on invocation context.

jfkonecn:
> [ai-generated] AI code review completed for this diff. No merge blockers found across query safety, logging, accessibility impact, or test updates. Non-blocking follow-up: consider adding a regression test for the duplicate-candidate query behavior when userId is null, since the native SQL predicate "u.id <> :userId" can evaluate to unknown and filter out all rows depending on invocation context.

I don't think this is a realistic issue.

src/main/java/org/mskcc/cbio/oncokb/repository/projection/PotentialDuplicateUserProjection.java
```diff
@@ -0,0 +1,17 @@
+package org.mskcc.cbio.oncokb.repository.projection;
```
bprize15, line 1: Nit: why are we separating into interface and separate impl class?

bprize15 (approved):
LGTM, one minor nit

src/main/java/org/mskcc/cbio/oncokb/repository/projection/PotentialDuplicateUserProjection.java
jfkonecn: The impl class is used for testing. Java to magically create a class for you when you do the query. 🤷 
I'm glad we're moving to go.

PR #1412, opened 2026-09-10
Fix: add trial account lifecycle with user_trial and trial_status

jfkonecn:
## Summary
- Adds a first-class trial account lifecycle with `user_trial` data and `trial_status` tracking.
- Aligns backend and UI flows for `REGULAR`, `TRIAL_PENDING_TERMS_ACCEPTANCE`, and `TRIAL`.
- Includes follow-up test coverage and migration compatibility hardening for the same scope.

## Problem
- Trial state transitions and activation flow were not fully modeled or consistently enforced across backend auth/service/resource paths.

## Root cause
- Missing persisted trial lifecycle structure and incomplete transition/auth coverage led to inconsistent behavior risk.

## Fix approach
- Introduces `UserTrial` + `TrialStatus` model, repository/service/DTO/mapper wiring, and Liquibase migration.
- Updates login and account lifecycle handling in user/auth resource and service paths.
- Updates frontend user-management surfaces and generated API types for trial status visibility/actions.
- Adds targeted backend tests for token-auth branches and trial transition/convert flows.

- `./mvnw -Denforcer.skip=true -Dtest=UserAuthenticationTokenServiceTest test` -> PASS (4 tests)
- `./mvnw -Denforcer.skip=true -Dtest=DomainUserDetailsServiceIT test` -> PASS (16 tests)
- `./mvnw -Denforcer.skip=true -Dtest=UserServiceIT#assertThatConvertUserToRegularClearsPendingTrialData,UserServiceIT#assertThatConvertUserToRegularKeepsActivatedTrialHistory,UserResourceIT#updateUserToPendingTrialInitiatesTrialActivation,UserResourceIT#updateUserToTrialApprovesTrialUser,UserResourceIT#updateUserToRegularConvertsTrialUser test` -> PASS (5 tests)

## Risk / Rollout
- Main risk is behavior regression in account lifecycle transitions; mitigated by targeted integration/unit coverage.
- Frontend auto-redirect behavior for trial terms is intentionally not included in this PR (kept reverted) to limit scope.

## Related issues
- Refs oncokb/oncokb-pipeline#1254


jfkonecn (commented):
[ai-generated] Thanks for the substantial trial-lifecycle refactor. I found one blocker and a few high-value follow-ups:\n\n1) **Blocker - N+1 query risk introduced in mapper**\n now does  for every mapped user. This path is used in list/hydration flows (e.g., user management pages), so it can add one DB query per row and regress admin-page latency under large result sets.\n\n   Suggested fix: batch-fetch  records for the page/user-id set (or join-fetch/projection) and attach by  map, similar to the existing bulk hydration strategy already used for .\n\n2) **Logging follow-up (non-blocking)**\n now emits multiple  lines with per-user email/id context. This can be noisy and increases PII at info level. Consider keeping a single milestone  and downgrading per-step/user-detail logs to .\n\n3) **Commenting follow-up (non-blocking)**\nThe  trial transition block is behaviorally dense (state transition + side effects for email/token lifecycle). A short comment documenting the intended transition precedence () and why transitions happen after base update would help future maintainers avoid subtle regressions.

jfkonecn (commented):
[ai-generated] Thanks for the substantial trial-lifecycle refactor. I found one blocker and a few high-value follow-ups:

1) **Blocker - N+1 query risk introduced in mapper**
`UserMapper.userToUserDTO(User, UserDetails)` now does `userTrialRepository.findOneByUser(user)` for every mapped user. This path is used in list and hydration flows (for example user management pages), so it can add one DB query per row and regress admin-page latency under large result sets.

   Suggested fix: batch-fetch `user_trial` records for the page and user-id set (or join-fetch or projection) and attach by `userId` map, similar to the existing bulk hydration strategy already used for `UserDetails`.

2) **Logging follow-up (non-blocking)**
`SlackService.sendUserRegistrationToChannel` now emits multiple `INFO` lines with per-user email and id context. This can be noisy and increases PII at info level. Consider keeping a single milestone `INFO` and downgrading per-step user-detail logs to `DEBUG`.

3) **Commenting follow-up (non-blocking)**
The `UserResource.updateUser` trial transition block is behaviorally dense (state transition plus side effects for email and token lifecycle). A short comment documenting the intended transition precedence (`PENDING_TERMS -> TRIAL -> REGULAR`) and why transitions happen after the base update would help future maintainers avoid subtle regressions.

src/main/java/org/mskcc/cbio/oncokb/web/rest/UserResource.java
```diff
@@ -174,12 +176,39 @@ public ResponseEntity<UserDTO> updateUser(
         }
 
         Optional<UserDTO> updatedUser;
+        TrialStatus requestedTrialStatus = userDTO.getTrialStatus() == null ? TrialStatus.REGULAR : userDTO.getTrialStatus();
```
calvinlu3, line 179: If there is no trialStatus, I think it shouldn't default to regular and just be a no op? Since trialStatus is not a required field, then any frontend code could forget to pass it and then a trial could become a regular user. Could be even easier to miss if you just call the api endpoint directly via a script.

calvinlu3 (changes requested):
Product review:
1. If a regular user gets converted to a trial via the admin radio button or the modal, then no user_trial row is created and the user gets `403 FORBIDDEN "Trial activation is incomplete. Please use your activation link again or contact support."` error. There is trial email sent to user.
2. On admin page, I'm not sure if its a valid option to manually change someone's Account Type to `pending trial terms acceptance`.
3. Users who at some point was on a trial license (now on regular) will have still have `trialAccount` in the additionalInfo JSON and the liquibase migration script currently makes their account Trial again. (ie id 2253)
4. Some trial users in prod got changed to trial after liquibase migration. (ie id 17456).
5.

PR #1415, opened 2026-09-15
show resistance description on public website

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/1268


src/main/webapp/app/shared/api/generated/OncoKbAPI-docs.json
```diff
@@ -4152,6 +4152,15 @@
         "query": {
           "$ref": "#/definitions/Query"
         },
+        "resistanceDescription": {
+          "type": "string",
+          "description": "(Nullable) The highest sensitivity level from a list of therapeutic evidences.",
```
calvinlu3, line 4157: Just noticed that this is not the right description, so probably have to change in core.

calvinlu3 (changes requested):
one comment that needs to be address in core

calvinlu3 (approved):

